PR #51, opened 2019-02-02
[MAINT] Update to keep up with poldracklab/niworkflows#299

oesteban:


smriprep/cli/run.py
```diff
@@ -505,7 +505,8 @@ def build_workflow(opts, retval):
     # First check that bids_dir looks like a BIDS folder
     bids_dir = os.path.abspath(opts.bids_dir)
     subject_list = collect_participants(
-        bids_dir, participant_label=opts.participant_label)
+        bids_dir, participant_label=opts.participant_label,
```
effigies: Did you want to create a layout here:

```Python
layout = BIDSLayout(bids_dir)
subject_list = collect_participants(layout, opts.participant_label)
```

smriprep/workflows/base.py
```diff
@@ -347,7 +347,8 @@ def init_single_subject_wf(subject_id, name, reportlets_dir, output_dir, bids_di
             't1w': ['/completely/made/up/path/sub-01_T1w.nii.gz'],
         }
     else:
-        subject_data = collect_data(bids_dir, subject_id)[0]
+        subject_data = collect_data(bids_dir, subject_id,
+                                    bids_validate=False)[0]
```
effigies: And then pass the layout to `init_single_subject_wf()` so you can do here:

```Python
subject_data = collect_data(layout, subject_id)[0]

oesteban: Before going down the globals, do you think we may roll this back to 33d4930 and keep going with that for now?

oesteban: I'll try, but I think it needs to be pickleable for creating the workflow in an isolated process. The layout is not an input to BIDSInfo, it is a class attribute and I think they are not pickled.

effigies: Unless you're transferring the actual object through a file descriptor, there's no need to pickle.

oesteban: You are right, this works :+1:

PR #57, opened 2019-03-07
[ENH] Pure Nipype brain extraction workflow

oesteban:
Using the implementation of ``antsBrainExtraction.sh`` from niworkflows
(supersedes #48)

### DIFF
--- a/smriprep/workflows/anatomical.py
+++ b/smriprep/workflows/anatomical.py
@@ -19,7 +19,9 @@ from nipype.interfaces import (
     fsl,
     image,
 )
-from nipype.interfaces.ants import BrainExtraction, N4BiasFieldCorrection
+
+from nipype.interfaces.ants.base import Info as ANTsInfo
+from nipype.interfaces.ants import N4BiasFieldCorrection
 from templateflow.api import get as get_template, get_metadata
 
 from niworkflows.engine.workflows import LiterateWorkflow as Workflow
@@ -33,6 +35,7 @@ from niworkflows.interfaces.freesurfer import (
 )
 from niworkflows.interfaces.images import TemplateDimensions, Conform
 from niworkflows.utils.misc import fix_multi_T1w_source_name, add_suffix
+from niworkflows.anat.ants import init_brain_extraction_wf
 from .outputs import init_anat_reports_wf, init_anat_derivatives_wf
 from .surfaces import init_surface_recon_wf
 
@@ -186,7 +189,7 @@ white-matter (WM) and gray-matter (GM) was performed on
 the brain-extracted T1w using `fast` [FSL {fsl_ver}, RRID:SCR_002823,
 @fsl_fast].
 """.format(
-        ants_ver=BrainExtraction().version or '<ver>',
+        ants_ver=ANTsInfo.version() or '<ver>',
         fsl_ver=fsl.FAST().version or '<ver>',
         template_name=template_meta['Name'],
         template_refs=', '.join(template_refs),
@@ -208,7 +211,7 @@ with `N4BiasFieldCorrection` [@n4], distributed with ANTs {ants_ver} \
 
     workflow.__desc__ = desc.format(
         num_t1w=num_t1w,
-        ants_ver=BrainExtraction().version or '<ver>'
+        ants_ver=ANTsInfo.version() or '<ver>'
     )
 
     inputnode = pe.Node(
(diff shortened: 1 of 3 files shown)


smriprep/workflows/anatomical.py
```diff
@@ -209,11 +209,18 @@ def init_anat_preproc_wf(skull_strip_template, fs_spaces, template, debug,
 [@ants, RRID:SCR_004757]"""
     desc += '.\n' if num_t1w > 1 else ", and used as T1w-reference throughout the workflow.\n"
 
+    desc = """\
+The T1w-reference was then skull-stripped with a *Nipype* implementation of
+the `antsBrainExtraction.sh` workflow (from ANTs), using {skullstrip_tpl}
+as target template.
+""".format(skullstrip_tpl=skull_strip_template)
```
effigies: What's our minimum Python? If it's 3.6 or higher, we can switch to [f-strings](https://docs.python.org/3/reference/lexical_analysis.html#formatted-string-literals).

smriprep/workflows/anatomical.py
```diff
@@ -231,15 +241,18 @@ def init_anat_preproc_wf(skull_strip_template, fs_spaces, template, debug,
 
     # 3. Skull-stripping
     # Bias field correction is handled in skull strip workflows.
-    skullstrip_ants_wf = init_skullstrip_ants_wf(name='skullstrip_ants_wf',
-                                                 skull_strip_template=skull_strip_template,
-                                                 debug=debug,
-                                                 omp_nthreads=omp_nthreads)
+    bran_extraction_wf = init_brain_extraction_wf(
```
effigies:
```suggestion
    brain_extraction_wf = init_brain_extraction_wf(
```

smriprep/workflows/anatomical.py
oesteban: I was actually thinking of that as well. We have Python +3.5 but I'd seriously consider enforcing +3.6 for that very reason.

smriprep/workflows/anatomical.py
oesteban: OMG never push a commit while you hear your kid crying :(

oesteban:
Before: https://256-160736148-gh.circle-artifacts.com/0/tmp/ds054/derivatives/smriprep/sub-100185.html
After: https://213-164755300-gh.circle-artifacts.com/0/tmp/ds054/derivatives/smriprep/sub-100185.html

effigies:
> Note: none of the builds are executed with fixed seed for atropos, hence the little differences.

Is it possible to fix the seed by environment variable?

oesteban:
We can include that. Are you thinking of a shared env variable for all algorithms accepting a seed?

Something like `$BIDS_PREP_SEED` ?

effigies:
No, I meant more "Does Atropos accept a seed"? I would rather we let these things be controlled by the underlying tools if at all possible, just to keep things simple.

effigies:
I suppose I should say this shouldn't hold up this PR. I think we should be deliberate and perhaps solicit input from @repronim before settling on an approach, anyway. (Also, it looks like ANTs does have some work in this direction: https://github.com/ANTsX/ANTs/pull/599)

oesteban:
[Atropos does not accept a seed, it only provides a boolean switch to make it random or not](https://github.com/ANTsX/ANTs/blob/15f4e4013ed33b9d226c8d1b7d9509b2a8b19ba2/Examples/Atropos.cxx#L760)

effigies:
I wonder if we can persuade them to move in a more consistent direction with the Registration stuff.

oesteban:
Alright, is it okay to merge then?

smriprep/workflows/anatomical.py
```diff
@@ -258,12 +270,12 @@ def init_anat_preproc_wf(skull_strip_template, fs_spaces, template, debug,
...
                 ('subject_id', 'inputnode.subject_id')]),
             (anat_template_wf, surface_recon_wf, [('outputnode.t1_template', 'inputnode.t1w')]),
-            (skullstrip_ants_wf, surface_recon_wf, [
-                ('outputnode.out_file', 'inputnode.skullstripped_t1'),
-                ('outputnode.out_segs', 'inputnode.ants_segs'),
-                ('outputnode.bias_corrected', 'inputnode.corrected_t1')]),
-            (skullstrip_ants_wf, applyrefined, [
-                ('outputnode.bias_corrected', 'in_file')]),
+            (brain_extraction_wf, surface_recon_wf, [
+                (('outputnode.bias_corrected', _pop), 'inputnode.skullstripped_t1'),
```
effigies: Bias corrected isn't the same as skull stripped, is it?

oesteban: Yes, it was. This is one of the reasons why I wanted to translate this workflow into nipype. In a future PR I want to address these:

- Bias-corrected should not be the same as skull stripped
- Add an additional N4 pass on the original image (not truncated, as it is now) with the brain mask, so both bias-corrected and skull-stripped have this more accurate correction.
- Reuse the initial transforms in the normalization workflow if the template used is the same (and consequently move to MNI152NLin2009cAsym as target template for skull-stripping).

effigies: But we save the bias-corrected file as the t1_preproc (L253), and that's not skull-stripped.

oesteban: You are right. It should be fixed now.

PR #81, opened 2019-05-02
ENH: Gauging whether ``updatehash=True`` is a good idea

oesteban:


oesteban:
Doesn't look like a great idea for now.

PR #99, opened 2019-07-12
CI: Fix PyPI deployment

effigies:
PyPI deployment failed. Copying the deployment test and just adding a `twine upload` step.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -170,14 +170,11 @@ jobs:
               echo "Skipping pytest job"
               circleci step halt
             fi
-      - attach_workspace:
-          at: /tmp
       - run:
           name: Check PyPi preconditions
           command: |
             pyenv global 3.5.2
             pip install --upgrade "setuptools>=30.3.0" "pip>=18.1" twine docutils
-            python setup.py check -r -s
             python setup.py sdist
             twine check dist/*.tar.gz
       - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -252,23 +249,20 @@ jobs:
...
+            pip install --upgrade "setuptools>=30.3.0" "pip>=18.1" twine docutils
             python setup.py sdist
-            twine upload dist/*
-            cd wrapper && python setup.py sdist
-            twine upload dist/*
+            twine check dist/*.tar.gz
+      - run:
+          name: Deploy to PyPi
+          command: |
+            twine upload dist/*.tar.gz
```
oesteban: Isn't this necessary?
```suggestion
            pyenv global 3.5.2
            pip install twine
            twine upload dist/*.tar.gz
```

effigies: No. https://github.com/poldracklab/niworkflows/blob/3ed92ea3b89189c4a86a41d99f6c95eae8829497/.circleci/config.yml#L220-L248

.circleci/config.yml
```diff
@@ -170,14 +170,14 @@ jobs:
               echo "Skipping pytest job"
               circleci step halt
             fi
-      - attach_workspace:
-          at: /tmp
       - run:
-          name: Check PyPi preconditions
+          name: Setup Python environment
           command: |
             pyenv global 3.5.2
```
effigies: I split this out in the earlier job so that it is obvious it will work later.

PR #102, opened 2019-09-03
FIX: Render INU-corrected T1w in Segmentation reportlet

### LINKED ISSUE #101: Segmentation and normalization reportlets show the INU-uncorrected T1w image
Comes from https://neurostars.org/t/fmriprep-v1-4-0-does-not-correct-for-inu-and-generate-a-full-html-report/4439/2?u=oesteban

oesteban:
Fixes #101.

### DIFF
--- a/smriprep/workflows/anatomical.py
+++ b/smriprep/workflows/anatomical.py
@@ -335,8 +335,8 @@ the brain-extracted T1w using `fast` [FSL {fsl_ver}, RRID:SCR_002823,
             (('t1w', fix_multi_T1w_source_name), 'inputnode.source_file')]),
         (anat_template_wf, anat_reports_wf, [
             ('outputnode.out_report', 'inputnode.t1_conform_report')]),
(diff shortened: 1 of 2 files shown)


smriprep/workflows/anatomical.py
```diff
@@ -335,8 +335,8 @@ def init_anat_preproc_wf(
             (('t1w', fix_multi_T1w_source_name), 'inputnode.source_file')]),
         (anat_template_wf, anat_reports_wf, [
             ('outputnode.out_report', 'inputnode.t1_conform_report')]),
-        (anat_template_wf, seg_rpt, [
-            ('outputnode.t1_template', 'in_file')]),
+        (brain_extraction_wf, seg_rpt, [
+            (('outputnode.bias_corrected', _pop), 'in_file')]),
```
effigies: Could you clarify why `_pop` is needed here? Shouldn't there only be one bias-corrected file?

PR #113, opened 2019-09-17
ENH: Write out the fsnative-to-T1w transform

oesteban:
Add a datasink to include a `_from-fsnative_to-T1w_mode-image_xfm.txt`
file that maps coordinates from the original T1w image into from FS'
conformed space coordinates.

Another improvement to ease #107 .


oesteban:
@effigies any obvious argument against this that I'm not seeing?

effigies:
Oh sorry, I saw this last night, and but didn't mark it unread, so it fell off my plate. Give me three minutes.

Issue #27, opened 2018-07-02
Make reusing freesurfer outputs more explicit


chrisgorgo:
Reusing precomputed freesurfer outputs is currently supported, but requires a very specific approach that is often confusing for users. To make things more transparent I propose adding `--reuse-freesurfer-dir` flag pointing to a SUBJECTS_DIR which will be reused. When that flag is set recon-all should not be attempted.

oesteban:
I guess the only missing bit here is to propagate the `--fs-subjects-dir` flag to fMRIPrep?

effigies:
I think that's right.

oesteban:
Opened a particular issue for this in fMRIPrep proper.

PR #114, opened 2019-10-07
ENH: Add --fs-subjects-dir flag

effigies:
Working toward #27.

### DIFF
--- a/smriprep/workflows/base.py
+++ b/smriprep/workflows/base.py
@@ -29,6 +29,7 @@ from .anatomical import init_anat_preproc_wf
 def init_smriprep_wf(
     debug,
     freesurfer,
+    fs_subjects_dir,
     hires,
     layout,
     longitudinal,
@@ -60,6 +61,7 @@ def init_smriprep_wf(
         wf = init_smriprep_wf(
             debug=False,
             freesurfer=True,
+            fs_subjects_dir=None,
             hires=True,
             layout=BIDSLayout('.'),
             longitudinal=False,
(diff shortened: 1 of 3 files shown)


smriprep/workflows/base.py
```diff
@@ -129,6 +133,8 @@ def init_smriprep_wf(
                 spaces=[s for s in output_spaces.keys() if s.startswith('fsaverage')] + [
                     'fsnative'] * ('fsnative' in output_spaces)),
             name='fsdir_run_%s' % run_uuid.replace('-', '_'), run_without_submitting=True)
+        if fs_subjects_dir is not None:
+            fsdir.inputs.subjects_dir = str(fs_subjects_dir.absolute())
```
oesteban: Is `absolute()` better than `resolve()` for some reason? (this is a question, not a change suggestion here).

oesteban:
Is poldracklab/niworkflows#398 released? Do you need to update some pin version?

smriprep/workflows/base.py
effigies: It doesn't attempt to resolve symlinks, just turns relative paths into absolute.

effigies:
> Is [poldracklab/niworkflows#398](https://github.com/poldracklab/niworkflows/pull/398) released? Do you need to update some pin version?

Good catch. Thanks.

smriprep/workflows/base.py
oesteban: Thanks for the clarification - pretty useful.

oesteban:
Do you want to test this new flag on circle? (i.e., update the ds005 build to exercise it)

oesteban:
I think a new section under Usage or Q in the FAQ regarding the reuse of FreeSurfer derivatives could be a great improvement. Also, some updates to https://fmriprep.readthedocs.io/en/latest/outputs.html#freesurfer-derivatives could be good to have.

Probably just linking https://github.com/poldracklab/fmriprep/issues/1573 would satisfy my request. Merging.

effigies:
Oh sure. When we add this to fMRIPrep, we should definitely update those docs. Just not much to do here in sMRIPrep that I can see.

Issue #1, opened 2019-01-08
Create a spatial normalization workflow


oesteban:
Transferred from https://github.com/poldracklab/fmriprep/pull/1115

PR #116, opened 2019-10-07
Skip brain extraction 

### LINKED ISSUE #12: Gracefully handle pre-skull-stripped T1w images
A number of subjects in ds000223 (e.g. [sub-07](https://s3.amazonaws.com/openneuro/ds000223/ds000223_R1.0.0/uncompressed/derivatives/mriqc/sub-07_T1w.html)) have pre-skull-stripped T1w images. ANTs skull-stripping on these datasets clips most superficial cortex, resulting in very poor preprocessing.

I propose adding a `--pre-skull-stripped` flag (or similar) to indicate that the anatomical workflow should assume the incoming image is already skull-stripped.

Alternately, we could do a test for a given percentage of zero voxels as evidence for skull-stripping, though this would need to be tested against defacers, as well.

adelavega:
Fixes #12 by creating an alternative workflow to `init_brain_extraction_wf`, `init_n4_only_wf`.
This wf only does n4 bias field correction, and creates a mask file by thresholding the input T1w image.

My main concern is if this is the right approach. 
Alternatively, the `init_brain_extraction_wf` workflow could be altered in niwokflows, but as this is supposed to be an implementation of ANTS' brain extraction, I thought it was better not to touch that.

I also though about just changing the `init_anat_preproc_wf` function to change the connections based on if brain extraction would be run, but that got messy quick.

Also, I'm currently just using a parameter to indicate if brain extraction should be skipped. Although the function to detect if brain extraction is needed is easy, I'm not sure how to dynamically create a workflow based on the output of a Node. 

Any pointers here?


smriprep/workflows/anatomical.py
```diff
@@ -27,18 +29,84 @@
...
+    inputnode = pe.Node(niu.IdentityInterface(fields=['in_files', 'in_mask']),
+                        name='inputnode')
+
+    outputnode = pe.Node(niu.IdentityInterface(
+        fields=['out_file', 'out_mask', 'bias_corrected', 'bias_image',
+                'out_segm', 'out_tpms']),
+        name='outputnode')
+
+    # Create brain mask (out_mask)
+    thr_brainmask = pe.MapNode(ThresholdImage(
```
oesteban: I think you'd be much better off using [`niworkflows.interfaces.nibabel.Binarize`](https://github.com/poldracklab/niworkflows/blob/master/niworkflows/interfaces/nibabel.py#L27) here.

smriprep/workflows/anatomical.py
```diff
@@ -27,18 +29,84 @@
...
+    # INU correction (output_image -> bias_corrected, out_file)
+    inu_n4 = pe.MapNode(
+        N4BiasFieldCorrection(
+            dimension=3, save_bias=True, copy_header=True,
+            n_iterations=[50] * 5, convergence_threshold=1e-7, shrink_factor=4,
+            bspline_fitting_distance=200),
+        n_procs=omp_nthreads, name='inu_n4_final', iterfield=['input_image'])
+
+    # Tolerate missing ANTs at construction time
+    _ants_version = Registration().version
```
oesteban: I believe this should work out:
```suggestion
    _ants_version = N4BiasFieldCorrection().version
```

smriprep/workflows/anatomical.py
```diff
@@ -27,18 +29,84 @@
...
+
+    if _ants_version and parseversion(_ants_version) >= Version('2.1.0'):
```
oesteban: I'm thinking a `try .. except ValueError` would be much more pythonic (that applies to existing code too, so leave it if you don't like the suggestion)

smriprep/workflows/anatomical.py
```diff
@@ -27,18 +29,84 @@
...
+        inu_n4.inputs.rescale_intensities = True
+    else:
+        warn("""\
+Found ANTs version %s, which is too old. Please consider upgrading to 2.1.0 or \
+greater so that the --rescale-intensities option is available with \
+N4BiasFieldCorrection.""" % _ants_version, DeprecationWarning)
+
+    copy_xform = pe.Node(CopyXForm(
```
oesteban: If you use binarize as suggested above, is this still necessary?

adelavega: is `CopyXForm` necessary?

adelavega:
Thanks for taking a look.

1) Yes that's what I was thnking and what @effigies had suggested. It shouldn't in principle hurt to run that twice, whereas running brain extraction twice does hurt
2) Sure, I just wasn't sure if that was necessary, but that'd be easy enough. In that case though, would you rather keep this workflow here, or modify the ants one to just skip brain extraction given a flag? Again, I can understand how keeping that one strictly like the ants workflow makes sense so I have no strong opinion
3) Makes sense

Any suggestions re: the brain-mask detection though? Or are you happy to leave it as an explicit CLI option?

adelavega:
I successfully ran ds001110/SherlockMerlin with brain extraction turned off, and the results look decent as far as I can tell.

This is ready for another review whenever you get a chance. No rush from me as preprocessing that dataset was my main goal.

Here are the reports for the dataset if anyone cares to take a look: https://web.corral.tacc.utexas.edu/neuroscout-datasets/SherlockMerlin/fmriprep/

Main points left to discuss:

1. Should this behavior be handled through a flag or detection? Either way I think that should go into `fmriprep` and this is fine as is. If we do detection, I guess we need to do it outside of nipype using pybids + the function I posted before. (Unless anyone knows how to do a conditions graph)
2) Should this new workflow go here or in niworkflows? Up to you. I'm happy to open another PR there with just the init workflow function. 
3) The `disallow_premasked` flag only makes sense in context of detection, right? If so, I'll wait until we resolve #1 

cc: @effigies

adelavega:
One more thing: I noticed `CopyXForm` sneakily also does a "pop" off a list. Thus, I put it back in, partly because its used for those outputs in the other workflow, and thus might be necessary.

effigies:
Cool. Will have a look now. Should be able to fix the tests by merging/rebasing master.

effigies:
Haven't configured the PEP8 speaks... Ignore the line lengths.

effigies:
Also, something's come up, and I might not finish a review this afternoon.

adelavega:
update: I didnt realize `brain_extraction_wf` gets the average T1w as input. In that case, the thresh_low of `0` wasn't high enough and resulted in masks like:

image

I think this led to a series of errors which prevented confounds from being extracted. 

I set it to 10, and that looks pretty close to what you get by threshold a single T1w with 0. A treshold of 1 should also be acceptable. I have no clue what the range of values of T1w images tend to be. 

image


I'll re-run with these values.

adelavega:
I undid the merge master (at least temporarily), because it seems that `fmriprep` is lagging behind `smriprep` (specifically in the naming of the outputs that was recently done)

adelavega:
Allright, all looks good. The latest reports ran without error. The only minor and potentially unrelated question I have, is that I got a ton of `a_comp_cor` confounds (a_comp_cor_514!). Is that just indicate of the noise in the dataset? Or can it possibly have something to do with this PR?

adelavega:
All tests are passing aside from some unrelated errors (blas, and a connection error).
This is ready to review. Would be OK with me to merge this and leave the fmriprep PR open if that one needs more work.

effigies:
Can you merge master? Should fix both tests...

adelavega:
@effigies tests passing fully now

smriprep/workflows/anatomical.py
```diff
@@ -27,17 +28,194 @@
...
+                        name='inputnode')
+
+    outputnode = pe.Node(niu.IdentityInterface(
+        fields=['out_file', 'out_mask', 'bias_corrected', 'bias_image',
+                'out_segm', 'out_tpms']),
+        name='outputnode')
+
+    # Create brain mask
+    thr_brainmask = pe.Node(
+        Binarize(thresh_low=10), name='binarize')
```
effigies: This feels high, though I also don't have numbers here. For the image you show in https://github.com/poldracklab/smriprep/pull/116#issuecomment-541938810, what is the result of:

```Python
data = nb.load(fname).get_fdata()
np.max(data[data < 10])
```

If this is << 1 (this is my guess), I would much prefer that we use a small number such as 1 or even 0.1. That makes it clear that it's interpolation noise that we'll be masking out, not very dark CSF.

adelavega: I agree in principle, and I was hoping for some feedback as you have more expertise on what the range of values tend to be. 

But I think 0.1 is definitely too low, and 1 might still be. FWIW, at least on this image, all "real" values, even the darkest CSF were more like 100+.

I get:
```
In [6]: data = nb.load(fname).get_fdata()
   ...: np.max(data[data < 10])
   ...: 
Out[6]: 9.0
```

adelavega: If you try this with any value though you get a similar result. My guess is this is all happening near the edge of the brain, and most of the non-contiguous noise is maybe no higher than 5? It's hard to say from a single sample though. 

```
In [8]: data = nb.load(fname).get_fdata()
   ...: np.max(data[data < 20])
Out[8]: 19.0
```

effigies: Perhaps we can get a better sense of this noise with a histogram?

```Python
import matplotlib.pyplot as plt
plt.hist(data[data < 20], bins=20)
```

IDK. I don't really have a sense of what the lowest CSF T1w values we're likely to find will be. OTOH, we can also just put this out and let people complain if we rip holes in their ventricles, since this is already a kind of an unrecommended hack.

> I was hoping for some feedback as you have more expertise on what the range of values tend to be.

Since we're not using quantitative MRI and getting actual T1 values, the scale is arbitrary. I suspect the range varies a lot across scanner. And any preprocessing that was done to get to the skull-stripped images is obviously free to rescale.

adelavega: image
image

Okay looks like we could safely set the threshold much lower

adelavega: Maybe 1-2 is good.

adelavega: Yes. My thoughts on the check was that it could be done in fmriprep (or at least outside this wf) and in smriprep it could just be an off/on flag. That was the test can be outside of nipype and use pybids to quickly run it, and set the flag here. 

The test of checking in an individual image that all extreme corner values == 0 worked for me. It's just in the averaged image we are dealing with here that we start getting values slightly above 0.

adelavega:
@oesteban this is ready for another review by you

smriprep/workflows/anatomical.py
```diff
@@ -27,17 +28,194 @@
...
+            (thr_brainmask, atropos_wf, [
+                ('out_mask', 'inputnode.in_mask')]),
+            (thr_brainmask, atropos_wf, [
+                ('out_mask', 'inputnode.in_mask_dilated')]),  # Dilate?
+            (atropos_wf, sel_wm, [('outputnode.out_tpms', 'inlist')]),
+            (sel_wm, inu_n4_final, [('out', 'weight_image')]),
+            (atropos_wf, outputnode, [
+                ('outputnode.out_segm', 'out_segm'),
+                ('outputnode.out_tpms', 'out_tpms')]),
+        ])
```
oesteban: I'm not going to ask you to do that in this PR, but I think this could be better modularized in niworkflows: https://github.com/poldracklab/niworkflows/issues/424

smriprep/workflows/anatomical.py
```diff
@@ -27,17 +28,194 @@
...
+                        name='inputnode')
+
+    outputnode = pe.Node(niu.IdentityInterface(
+        fields=['out_file', 'out_mask', 'bias_corrected', 'bias_image',
+                'out_segm', 'out_tpms']),
+        name='outputnode')
+
+    # Create brain mask
+    thr_brainmask = pe.Node(
+        Binarize(thresh_low=2), name='binarize')
```
oesteban:
```suggestion
    thr_brainmask = pe.Node(Binarize(thresh_low=2), name='binarize')
```

smriprep/workflows/anatomical.py
```diff
@@ -27,17 +28,194 @@
...
+    # Tolerate missing ANTs at construction time
+    _ants_version = N4BiasFieldCorrection().version
+
+    if _ants_version and parseversion(_ants_version) >= Version('2.1.0'):
+        inu_n4_final.inputs.rescale_intensities = True
+    else:
+        warn("""\
+Found ANTs version %s, which is too old. Please consider upgrading to 2.1.0 or \
+greater so that the --rescale-intensities option is available with \
+N4BiasFieldCorrection.""" % _ants_version, DeprecationWarning)
```
oesteban:
```suggestion
    try:
        inu_n4_final.inputs.rescale_intensities = True
    except ValueError:
        warn("The installed ANTs version too old. Please consider upgrading to "
             "2.1.0 or greater.", DeprecationWarning)

File: smriprep/workflows/anatomical.py (hunk 4 above)
Comment (oesteban) [reply smriprep/workflows/anatomical.py]: please remove this one node :+1:

File: smriprep/workflows/anatomical.py [hunk 9]
```diff
@@ -27,17 +28,194 @@
...
 from niworkflows.utils.misc import fix_multi_T1w_source_name, add_suffix
-from niworkflows.anat.ants import init_brain_extraction_wf
+from niworkflows.anat.ants import (
+    init_brain_extraction_wf, init_atropos_wf, ATROPOS_MODELS, _pop
+    )
 from .norm import init_anat_norm_wf
 from .outputs import init_anat_reports_wf, init_anat_derivatives_wf
 from .surfaces import init_surface_recon_wf
 
+from packaging.version import parse as parseversion, Version
```
oesteban: This can go away if you accept my suggestion

smriprep/workflows/anatomical.py
adelavega: Yes, agreed

smriprep/workflows/anatomical.py
adelavega: I tried without it, but had to restore it for two outputs from `inu_n4_final`. 
Looks like the original brain extraction workflow also does this, even though `copy_header=True`. 
I can try again without but I remember it not working.

adelavega:
Sounds good. 
So do you want a new PR over in niworkflows or should I just clean this up?

smriprep/workflows/anatomical.py
```diff
@@ -27,17 +28,194 @@
...
+    **Parameters**
+
+        omp_nthreads : int
+            Maximum number of threads an individual process may use
+        mem_gb : float
+            Estimated peak memory consumption of the most hungry nodes
+        bids_suffix : str
+            Sequence type of the first input image. For a list of acceptable values
+            see https://bids-specification.readthedocs.io/en/latest/\
+04-modality-specific-files/01-magnetic-resonance-imaging-data.html#anatomy-imaging-data
```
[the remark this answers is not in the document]
oesteban: sphinx chokes with urls broken in two lines.

adelavega:
That makes sense to me. I'll have to deprioritize this for this and next week but the week after that I'll check back and see if I can be of help :)

oesteban:
I'll come back to this on Monday. My apologies, thanks very much for you patience :)

smriprep/workflows/anatomical.py
```diff
@@ -27,17 +28,180 @@
...
+            List of input anatomical images to be bias corrected,
+            typically T1-weighted.
+            If a list of anatomical images is provided, subsequently
+            specified images are used during the segmentation process.
+            However, only the first image is used in the registration
+            of priors.
+            Our suggestion would be to specify the T1w as the first image.
+
+
+    **Outputs**
```
oesteban:
```suggestion
    Outputs
```

smriprep/workflows/anatomical.py
```diff
@@ -27,17 +28,180 @@
...
+
```
oesteban:
```suggestion
    -------
```

smriprep/workflows/anatomical.py
```diff
@@ -27,17 +28,180 @@
...
+
```
oesteban: Remove this newline

smriprep/workflows/anatomical.py
```diff
@@ -27,17 +28,180 @@
...
+        bias_corrected
+            Same as "out_file", provided for consistency with brain extraction
+        bias_image
+            The :abbr:`INU (intensity non-uniformity)` field estimated for each
+            input in ``in_files``
+        out_segm
+            Output segmentation by ATROPOS
+        out_tpms
+            Output :abbr:`TPMs (tissue probability maps)` by ATROPOS
+
```
oesteban: Remove one newline

smriprep/workflows/anatomical.py
```diff
@@ -27,17 +28,180 @@
...
+        in_files
+            List of input anatomical images to be bias corrected,
+            typically T1-weighted.
+            If a list of anatomical images is provided, subsequently
+            specified images are used during the segmentation process.
+            However, only the first image is used in the registration
+            of priors.
+            Our suggestion would be to specify the T1w as the first image.
+
```
oesteban: Remove one newline

smriprep/workflows/anatomical.py
```diff
@@ -27,17 +28,180 @@
...
+            Whether ATROPOS should generate a random seed based on the
+            system's clock
+        atropos_model : tuple or None
+            Allows to specify a particular segmentation model, overwriting
+            the defaults based on ``bids_suffix``
+        name : str, optional
+            Workflow name (default: antsBrainExtraction)
+
+
+    **Inputs**
```
oesteban:
```suggestion
    Inputs
```

smriprep/workflows/anatomical.py
```diff
@@ -27,17 +28,180 @@
...
+
```
oesteban:
```suggestion
    ------
```

smriprep/workflows/anatomical.py
```diff
@@ -27,17 +28,180 @@
...
+    .. workflow::
+        :graph2use: orig
+        :simple_form: yes
+
+        from smriprep.workflows.anatomical import init_n4_only_wf
+        wf = init_n4_only_wf()
+
+
+    **Parameters**
```
oesteban:
```suggestion
    Parameters
```

smriprep/workflows/anatomical.py
```diff
@@ -27,17 +28,180 @@
...
+
```
oesteban:
```suggestion
    ----------
```

smriprep/workflows/anatomical.py
```diff
@@ -27,17 +28,180 @@
...
+        atropos_refine : bool
+            Enables or disables the whole ATROPOS sub-workflow
+        atropos_use_random_seed : bool
+            Whether ATROPOS should generate a random seed based on the
+            system's clock
+        atropos_model : tuple or None
+            Allows to specify a particular segmentation model, overwriting
+            the defaults based on ``bids_suffix``
+        name : str, optional
+            Workflow name (default: antsBrainExtraction)
```
oesteban: Please remove one indentation

smriprep/workflows/anatomical.py
```diff
@@ -27,17 +28,180 @@
...
+    **Inputs**
+
+        in_files
+            List of input anatomical images to be bias corrected,
+            typically T1-weighted.
+            If a list of anatomical images is provided, subsequently
+            specified images are used during the segmentation process.
+            However, only the first image is used in the registration
+            of priors.
+            Our suggestion would be to specify the T1w as the first image.
```
oesteban: Please remove one indentation

smriprep/workflows/anatomical.py
```diff
@@ -27,17 +28,180 @@
...
+            Calculated brain mask
+        bias_corrected
+            Same as "out_file", provided for consistency with brain extraction
+        bias_image
+            The :abbr:`INU (intensity non-uniformity)` field estimated for each
+            input in ``in_files``
+        out_segm
+            Output segmentation by ATROPOS
+        out_tpms
+            Output :abbr:`TPMs (tissue probability maps)` by ATROPOS
```
oesteban: Please remove one indentation

smriprep/workflows/anatomical.py
```diff
@@ -27,17 +28,180 @@
...
+                    mem_gb=3.0,
+                    bids_suffix='T1w',
+                    atropos_refine=True,
+                    atropos_use_random_seed=True,
+                    atropos_model=None):
+    """
+    An alternative workflow to "init_brain_extraction_wf", for anatomical
+    images which have already been brain extracted.
+
+
```
oesteban: Remove one line

smriprep/workflows/anatomical.py
```diff
@@ -27,17 +28,180 @@
...
+      1. Creates brain mask assuming all zero voxels are outside the brain
+      2. Applies N4 bias field correction
+      3. (Optional) apply ATROPOS and massage its outputs
+      4. Use results from 3 to refine N4 bias field correction
+
+
```
oesteban:
```suggestion
    Workflow Graph
```

smriprep/workflows/anatomical.py
```diff
@@ -27,17 +28,180 @@
...
+    .. workflow::
+        :graph2use: orig
+        :simple_form: yes
+
+        from smriprep.workflows.anatomical import init_n4_only_wf
+        wf = init_n4_only_wf()
```
oesteban: Please add one indentation level

smriprep/workflows/anatomical.py
```diff
@@ -27,17 +28,180 @@
...
+
+
```
oesteban: Please remove one newline

smriprep/workflows/anatomical.py
```diff
@@ -27,17 +28,180 @@
...
+def init_n4_only_wf(name='n4_only_wf',
+                    omp_nthreads=None,
+                    mem_gb=3.0,
+                    bids_suffix='T1w',
+                    atropos_refine=True,
+                    atropos_use_random_seed=True,
+                    atropos_model=None):
+    """
+    An alternative workflow to "init_brain_extraction_wf", for anatomical
```
oesteban:
```suggestion
    Build a workflow to sidetrack brain extraction on skull-stripped datasets.

    An alternative workflow to "init_brain_extraction_wf", for anatomical
```

smriprep/workflows/anatomical.py
```diff
@@ -27,17 +28,180 @@
...
+)
 from .norm import init_anat_norm_wf
 from .outputs import init_anat_reports_wf, init_anat_derivatives_wf
 from .surfaces import init_surface_recon_wf
 
+from warnings import warn
+
+
+def init_n4_only_wf(name='n4_only_wf',
+                    omp_nthreads=None,
```
oesteban:
```suggestion
                    omp_nthreads,
```
Please make this argument mandatory (will require writing name after)

smriprep/workflows/anatomical.py
```diff
@@ -27,17 +28,180 @@
...
+                    mem_gb=3.0,
+                    bids_suffix='T1w',
+                    atropos_refine=True,
+                    atropos_use_random_seed=True,
+                    atropos_model=None):
```
oesteban: Please sort arguments alphabetically

smriprep/workflows/anatomical.py
```diff
@@ -27,17 +28,180 @@
...
+            Output segmentation by ATROPOS
+        out_tpms
+            Output :abbr:`TPMs (tissue probability maps)` by ATROPOS
+
+
+    """
+    wf = pe.Workflow(name)
+
+    if omp_nthreads is None or omp_nthreads < 1:
+        omp_nthreads = cpu_count()
```
oesteban: Please remove

effigies:
Is this almost ready, or should we go ahead and release a bug-fix so we can get #146 in the next fMRIPrep?

adelavega:
If I recall, most of this should be a PR to niworkflows, so that requires that, plus a release there. 
Functionally though, it's pretty much ready.

adelavega:
I've made a PR to niworkflows: https://github.com/poldracklab/niworkflows/pull/435

I also added a function to check the extreme corners of an image for zero values, and automatically skip brain extraction if this is true for all T1ws. 

Things left to decide here:
- [ ]  If the function checking is in the right place. I think it more or less is, that way fmriprep doesnt have to do anything, and there's no duplication of that code
- [ ] Check CLI arguments. Maybe instead it should be `force_skull_strip`, in cases where our check think it should be skipped incorrectly. The behavior is already there, but the name.

This is all untested, btw.

setup.cfg
```diff
@@ -25,7 +25,7 @@ install_requires =
     matplotlib >= 2.2.0
     nibabel >= 2.2.1
     nipype >= 1.3.1
-    niworkflows ~= 1.0.0
+    niworkflows @ git+https://github.com/adelavega/niworkflows.git@2ce0b5b3ea04679788c17ccb84ad8a8ca2ef7fa6
```
effigies:
```suggestion
    niworkflows ~= 1.1.1
```

smriprep/workflows/anatomical.py
```diff
@@ -219,12 +227,33 @@ def init_anat_preproc_wf(
...
-        template_spec=skull_strip_template[1],
-        atropos_use_random_seed=not skull_strip_fixed_seed,
-        omp_nthreads=omp_nthreads,
-        normalization_quality='precise' if not debug else 'testing')
+    def _is_skull_stripped(imgs):
+        """ Checks that if T1w images are skulled stripped by interrogating
+        extreme planes for all-zeros
+        """
+        def _check_img(img):
+            data = img.get_data()
```
effigies:
```suggestion
            data = img.get_fdata(dtype=np.float32)
```

smriprep/workflows/anatomical.py
```diff
@@ -219,12 +227,33 @@ def init_anat_preproc_wf(
...
+            sidevals = data[0, :, :].sum() + data[-1, :, :].sum() + \
+                data[:, 0, :].sum() + data[:, -1, :].sum() + \
+                data[:, :, 0].sum() + data[:, :, -1].sum()
+            return sidevals > 0
```
effigies: Should consider the case where zeroes aren't exactly zero (e.g., float data saved as int16, so scaling parameters apply and push the 0 value slightly off):

```suggestion
            return sidevals > 10
```

Does that seem reasonably big to allow accumulation across 6 planes but well under what would happen if there were neck in one of them?

smriprep/workflows/anatomical.py
oesteban:
```suggestion
            data = np.abs(nb.load(img).get_fdata(dtype=np.float32))
```
Chris' recommendation plus weird in-scanner denoising + resampling that may introduce negative values.

smriprep/workflows/anatomical.py
oesteban: I might not be understanding the algorithm, but should the comparison be `sidevals < 10`?

My main concern would be:
  - skull-stripped images with the inferior plane cutting through the cerebellum (e.g., bad FoV prescription) would be false negatives here [although maybe you should just not use such bad scan].
  - images that have been rotated in defacing to be AC-PC aligned could potentially give false positives under certain circumstances (in-scanner denoising). Still, 10 should be sufficiently low.

adelavega: Oops, I think you're correct, it should be <10 (that is, the image is already skull stripped, and therefore should return True, and trigger n4 only workflow.

I think any heuristic is going to fail sometimes, which why I'd want to provide a CLI option to over-ride (should we allow both force skip, and force trip then?)

About #107, I'm not sure I understand. This was originally buit for a use case in which a raw BIDS dataset had skull stripped images (and were not labeled as such). This seems to be a not common enough usecase. 

I'm fine skipping the heuristic altogether though and just have a `skip-brain-extraction` CLI flag. Your call @oesteban

effigies:
What if we ask users to provide a brain mask? Then we just use it.

adelavega:
But then it would have to be per subject, and I think the brain mask extraction works pretty well (so why create more work for the user?)

Given this is fairly niche (although IMO important, given the # of openneuro datasets like this), I think a CLI flag might be a good and easy solution (especially since we would need an over-ride CLI flag anyways, so were adding yet another option no matter what).

smriprep/workflows/anatomical.py
oesteban: > About #107, I'm not sure I understand. This was originally buit for a use case in which a raw BIDS dataset had skull stripped images (and were not labeled as such). This seems to be a not common enough usecase.

Okay, sorry I misunderstood. This PR is to add a fall-back if the user (unknowingly) is passing in skull-stripped brains. Since the user doesn't know it, we can't expect them to place those brains in the derivatives folder.

In that case, I totally understand your approach - thanks for the patience. Let's go ahead with this. I'll push for the BIDS-validator to look into this particular, so it will give you a list of skull-stripped datasets, which you can move to a derivatives folder. Until that happens (if it ever happens), having this in fMRIPrep will be helpful :).

That said, all the proposals (current and alternatives) seem fine to me - my concern came from not understanding the problem addressed.

adelavega: Yep, you got it. Again, if at any point we think the heuristic is too unreliable, we can leave it as a CLI flag. 

That said, I think the current heuristic errs on the side of skull stripping (as the extreme values have to be very low to skip skull stripping automatically). So I think a false positive (i.e. non-skull stripped T1 detected as skull stripped) is fairly rare. 

Agree the BIDS-validator should check for this. In theory these datasets are BIDS-invalid, since they are not RAW. But in practice many such datasets exist in the wild. 

I fixed the logic in the test. Let me know if you need anything else.

adelavega:
Before we all forget about this PR, let me summarize:

- The `n4_only` workflow is merged and released on niworkflows
- The skull-strip detection seems to work fairly well, but we are concerned about edge cases

So we need to decide if s/fmriprep should either:

1) Detect skull stripped T1ws, with a flag to disable detection and `--force-skull-strip`. This is good if we think there will be false positives (non-stripped T1ws detected as pre-stripped).
2) The same as above, but instead of `force`, a `--skip-skull-strip` flag. This is good if we think there will be false negatives. 
3) A flag with the following options: `auto` `force` `skip`. This is the most flexible. 
4) No detection, only a `--skip-skull-strip` flag

Given that the detection could fail either way (although honestly, I think it's probably unlikely), I would vote for 3). 

I'm also fine with 4), as I think users with pre-stripped datasets may already know it (although I didn't at first, and I was puzzled as to why my results looked bad).

adelavega:
> DK if you want to preemptively merge/rebase on that, or try to squeeze this in and make @mgxd do the work.

Ahem, the latter if possible.

smriprep/workflows/base.py
```diff
@@ -329,7 +329,7 @@ def init_single_subject_wf(
         hires=hires,
         longitudinal=longitudinal,
         name="anat_preproc_wf",
-        num_t1w=len(subject_data['t1w']),
+        t1w=subject_data['t1w'],
```
effigies: You need to pass `skip_brain_extraction` here, and up the chain to the CLI.

effigies:
Need to update the niworkflows dependency here:

https://github.com/poldracklab/smriprep/blob/6a556bcf16a33ea21ed4fdcb8b7335332af10654/docs/requirements.txt#L5

effigies:
Pulling from a resolved comment:

> We also need to update `workflow.__desc__` to accurately reflect:
>
> 1. Skull-stripping is run
> 2. Skull-stripping is skipped
> 3. Pre-skull-stripped images were detected, and so skull-stripping was skipped.

smriprep/cli/run.py
```diff
@@ -95,6 +95,9 @@ def get_parser():
     g_ants.add_argument('--skull-strip-fixed-seed', action='store_true',
                         help='do not use a random seed for skull-stripping - will ensure '
                              'run-to-run replicability when used with --omp-nthreads 1')
+    g_ants.add_argument(
+        '--skip-brain-extraction', required=False, action='store_true',
+        default=False, help='Skip the brain extraction workflow (for pre-extracted T1ws))')
```
effigies: Hmm. This takes care of false negatives, but I suppose we should consider false positives, where we might incorrectly decide an image is skull-stripped.

Maybe we want `--skull-strip {auto,skip,force}`?

Or am I over-thinking the problem?

adelavega: Yeah, I previously suggested this, but also worried I was overthinking it. False positives are actually probably a bigger problem in a way (many more people have normal T1s)

If you're comfortable with the  additional CLI complexity, I think it this makes the most sense.

Again, we could also remove the detection if we think its not reliable, and only have a skip detection flag.

My preferences in order: 1) CLI argument w/ 3 options 2) skip only CLI flag, with no detection 3) skip flag + detection (current)

effigies: I lean toward 1 as well.

adelavega: @mgxd ? any opinion

smriprep/workflows/anatomical.py
```diff
@@ -22,15 +25,17 @@
 )
 from niworkflows.interfaces.images import TemplateDimensions, Conform, ValidateImage
 from niworkflows.utils.misc import fix_multi_T1w_source_name, add_suffix
-from niworkflows.anat.ants import init_brain_extraction_wf
+from niworkflows.anat.ants import (
+    init_brain_extraction_wf, init_n4_only_wf, _pop
```
mgxd: not a huge fan of importing "private" functions

docs/requirements.txt
```diff
@@ -2,12 +2,12 @@ git+https://github.com/AleksandarPetrov/napoleon.git@0dc3f28a309ad602be5f44a9049
...
 packaging
 pydot>=1.2.3
 pydotplus
 sphinx-argparse
 sphinx>=2.1.2
 sphinx_rtd_theme
 sphinxcontrib-apidoc ~= 0.3.0
-templateflow
\ No newline at end of file
+templateflow
```
mgxd: why is this different? spacing?

smriprep/cli/run.py
mgxd: i'm fine with the 3 way approach, with the default being `auto`. It should definitely document which route was taken

effigies:
Looks like @mgxd got in first. Merge/rebase?

smriprep/workflows/anatomical.py
adelavega: Sure, it's just a pop function so we can redefine.

docs/requirements.txt
adelavega: not sure 🤷

smriprep/cli/run.py
adelavega: Sure, how? You mean which option was select or what auto detected? And if so, where do I log that?

effigies:
Closed in favor of #167.

PR #119, opened 2019-10-11
CI: Test packaging and update deploy_pypi step

effigies:


effigies:
Looks like I didn't add the new step to the workflow...

PR #133, opened 2019-10-25
CI: Run pytests on Python 3.7 for now

effigies:
Scipy doesn't have a Python 3.8 wheel yet, and building it is a pain. Dropping pytests back to 3.7 for now.


effigies:
Fixed. Skipping the rest and merging.

PR #136, opened 2019-11-08
CI: Optimize CircleCI using a local docker registry instead docker save/load

oesteban:
Seems like replacing the docker save/load approach with a local registry run on the same machine allows us to make the main build job ~8min lighter and smoke tests builds ~6min lighter.

This is probably not a big deal time-wise, but I think this approach is more reliable than the docker save/load, and will escalate better if CircleCI opens up to having service jobs or some such.

This PR also revises the artifacts being collected - which were too many for successful runs (including the work directory and freesurfer folder).

The idea is to test how this works for sMRIPrep and then propagate to fMRIPrep and other projects if things look good.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -79,25 +79,15 @@ jobs:
           command: |
             docker exec -it registry /bin/registry garbage-collect --delete-untagged \
                 /etc/docker/registry/config.yml
-
-      - persist_to_workspace:
-          root: /tmp
-          paths:
-            - docker
-            - src/smriprep
-
-  update_cache:
-    machine:
-      # Ubuntu 14.04 with Docker 17.10.0-ce
-      image: circleci/classic:201711-01
-    working_directory: /tmp/src/smriprep
-    steps:
-      - attach_workspace:
-          at: /tmp
       - save_cache:
          key: build-v2-{{ .Branch }}-{{ epoch }}
          paths:
             - /tmp/docker
+
+      - persist_to_workspace:
+          root: /tmp
+          paths:
+            - src/smriprep
 
   get_data:
     machine:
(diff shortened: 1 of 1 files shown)


effigies:
I guess if it works, cool? I have no context for evaluating this apart from "Will the test pass".

oesteban:
Yup, it seems to work, looking at the layer digests and manifests. My only concern is that it seems to keep slowly growing. - I'm looking right now at how to garbage collect unused layers.

oesteban:
Nice - this https://stackoverflow.com/a/57989780 worked out locally. Testing now on Circle.

oesteban:
Working on Circle too! - https://app.circleci.com/jobs/github/oesteban/smriprep/941/parallel-runs/0/steps/0-107

The step to archive and propagate the workspace went back below 5m as in the early runs where the registry was still clean.

If tests pass, we can go ahead :)

oesteban:
Yass! Almost 1GB of difference after garbage collection:

- Last run (w/ cleaning up): [4.9GB / 2:54](https://app.circleci.com/jobs/github/oesteban/smriprep/946/parallel-runs/0/steps/0-101)
- Previous run (wo/ cleaning up): [5.8GB / 3:37](https://app.circleci.com/jobs/github/oesteban/smriprep/937/parallel-runs/0/steps/0-101)

.circleci/config.yml
```diff
@@ -59,30 +69,35 @@ jobs:
...
     working_directory: /tmp/src/smriprep
     steps:
       - attach_workspace:
           at: /tmp
       - save_cache:
-         key: build-v1-{{ .Branch }}-{{ epoch }}
+         key: build-v2-{{ .Branch }}-{{ epoch }}
          paths:
-            - /tmp/cache/docker.tar.gz
+            - /tmp/docker
```
effigies: In general I've found caches to be quicker, so there's no advantage to putting things in the workspace and then copying over to the cache in a separate job. Just purely use the cache.

effigies: See, e.g, https://github.com/poldracklab/fmriprep/pull/1608

oesteban:
Okay, the full workflow is now under 30 min! - Merging.

PR #138, opened 2019-11-09
FIX: Build smriprep-docker like fmriprep-docker

oesteban:
We were not running very strict tests, and that ended up submitting a version 99.99.99 to pypi.


oesteban:
Please pull my ears if I'm merging when I shouldn't. But I'd like to cut another rc and test out things.

PR #141, opened 2019-11-22
FIX: Allow setting nonstandard spaces for parser

oesteban:
Addresses [this problem](https://app.circleci.com/jobs/github/poldracklab/fmriprep/10357) - ref.: poldracklab/fmriprep#1827.

### DIFF
--- a/smriprep/cli/utils.py
+++ b/smriprep/cli/utils.py
@@ -4,6 +4,8 @@
 from argparse import Action
 from ..conf import TF_TEMPLATES as _TF_TEMPLATES
 
+EXCEPTIONS = None
+
 
 class ParseTemplates(Action):
     """Manipulate a string of templates with modifiers."""
@@ -27,6 +29,7 @@ def _template(inlist):
 
 def output_space(value):
     """Parse one element of ``--output-spaces``."""
+    global EXCEPTIONS
     tpl_args = value.split(':')
     template = tpl_args[0]
     spec = {}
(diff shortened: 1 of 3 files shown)


smriprep/cli/utils.py
```diff
@@ -27,16 +29,29 @@ def _template(inlist):
...
+
     if template not in _TF_TEMPLATES:
         raise ValueError("""\
 Template identifier "{}" not found. Please, make sure TemplateFlow is \
 correctly installed and contains the given template identifiers.""".format(template))
 
     return template, spec
+
+
+def set_nonstandard_spaces(inlist):
```
effigies: This function doesn't seem to be called anywhere.

oesteban: https://github.com/adelavega/fmriprep/commit/c7cddba027946cc8e9c1d524a98b2b855a032adc#diff-b93ab0122f4b2ac6738b896651cac6ccR53

effigies: Oh, I see...

Is there another way to do this? It feels very messy to fiddle with global state. What if we just passed a list of valid nonstandard spaces to `output_space`?

oesteban:
How would you pass it to ParseTemplate?

smriprep/cli/utils.py
oesteban: How would you pass it to `ParseTemplates`?

effigies: IDK... I'm not really aware of all this machinery. I'll look tomorrow to see if I can figure out what we're doing with CLI args these days and if there's something cleaner. If not, I guess do what you have to.

oesteban: For sure I can rework the action class... but I honestly took the lazy way for this.

oesteban: @effigies, more acceptable now?

PR #144, opened 2019-11-27
DOC: Refactor of documentation

oesteban:
- [x] Add a minimal structure
- [x] Rewrite API main document
- [x] Add CLI documentation (usage page)
- [x] Add an installation page
- [x] Add links to github repo for sources.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -21,9 +21,10 @@ from smriprep import (
     __version__ as _version,
     __copyright__ as _copyright,
 )
-from smriprep.__about__ import __url__ as github_url
 
 sys.path.append(os.path.abspath('sphinxext'))
+from github_link import make_linkcode_resolve
+
 sys.path.insert(0, os.path.abspath('../wrapper'))
 
 # -- Project information -----------------------------------------------------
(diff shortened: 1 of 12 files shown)


oesteban:
looking good - https://1034-164755300-gh.circle-artifacts.com/0/tmp/gh-pages/docs/_build/no_version_html/html/index.html

docs/conf.py
```diff
@@ -110,7 +99,9 @@ def linkcode_resolve(domain, info):
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
 # This pattern also affects html_static_path and html_extra_path.
-exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store', 'api/modules.rst']
+exclude_patterns = [
+    '_build', 'Thumbs.db', '.DS_Store',
+    'api/modules.rst', 'api/smriprep.rst']
```
oesteban: @rwblair I've found this easier than modifying apidoc templates to achieve a better rendering of the API (please check the corresponding ``docs/api.rst`` file).

rwblair: Makes sense, creating the api.rst directly is much more straightforward since its the only place the template is used and we aren't adding new submodules to be listed there that often.

PR #146, opened 2019-12-04
fix: empty specs for legacy/nonstd spaces

### LINKED ISSUE #145: Issue parsing templates
in the wild: poldracklab/fmriprep#1895

If no parameters are passed to template, default should be set to an empty dictionary, not None, to allow for downstream parameter querying.

mgxd:
Closes #145


effigies:
LGTM, but update the tests?

PR #148, opened 2019-12-08
FIX: Use T2/FLAIR refinement at cortribbon stage

effigies:
Reported in [Neurostars #5564](https://neurostars.org/t/autorecon-surfs0-mrisread-rh-white-could-not-open-file/5564), T2 processing in FreeSurfer runs into a data dependency failure because we're running `-T2pial` in `-autorecon-hemi` stage, while it needs to run immediately before `-cortribbon`.

It's pretty coarse. There's a part that has to be run serially, and then a part that can be parallelized across hemispheres. By using `-parallel`, we can get the parallelism, at the cost of 2x cores for the whole process. If nothing in [`-T2pial`, `-surfvolume` or `-cortribbon`](https://github.com/freesurfer/freesurfer/blob/2beb96c6099d96508246c14a24136863124566a3/scripts/recon-all#L4147-L4391) can take advantage of OpenMP parallelism, then this will be a 2-core job, which seems acceptable. If something can, it will probably be worth comparing to see which provides the better speedup.

### DIFF
--- a/smriprep/workflows/surfaces.py
+++ b/smriprep/workflows/surfaces.py
@@ -340,8 +340,13 @@ def init_autorecon_resume_wf(omp_nthreads, name='autorecon_resume_wf'):
     autorecon_surfs.interface._always_run = True
 
     # -cortribbon is a prerequisite for -parcstats, -parcstats2, -parcstats3
-    cortribbon = pe.Node(ReconAll(directive=Undefined,
-                                  steps=['cortribbon']), name='cortribbon')
+    # Claiming two threads because pial refinement can be split by hemisphere
+    # if -T2pial or -FLAIRpial is enabled.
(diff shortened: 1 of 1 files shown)


effigies:
Need to test it. I don't think any of our online tests use T2w.

effigies:
Result of test:

1) It works.
2) Some steps can take advantage of multiple threads. This requires setting the number of threads unconditionally, as T2/FLAIR presence is not known in advance, which will cause `cortribbon` to be a larger bottleneck.
3) I have not yet tested whether `-parallel` and claiming 2 threads would provide a comparable speedup while leaving more cores open for other jobs.

effigies:
Tested with `-parallel` instead of `-openmp 7`, and got a 30 minute speedup. This is good to merge, IMO.

PR #157, opened 2020-01-17
MAINT: Revise CircleCI to optimize TemplateFlow and caching

oesteban:
Also, simplify the Dockerfile accordingly

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -125,8 +125,7 @@ jobs:
 
   get_data:
     machine:
-      # Ubuntu 14.04 with Docker 17.10.0-ce
-      image: circleci/classic:201711-01
+      image: circleci/python:3.7.4
     working_directory: /home/circleci/data
     steps:
       - restore_cache:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -408,7 +437,8 @@ jobs:
                 -v /tmp/data:/tmp/data:ro \
                 -v /tmp/fslicense:/tmp/fslicense:ro \
                 -v /tmp/ds005:/tmp/ds005 \
-                -v /tmp/nipype.cfg:/home/smriprep/.nipype/nipype.cfg \
+                -v /tmp/templateflow:/home/smriprep/.cache/templateflow \
+                -v /tmp/src/smriprep/.circleci/nipype.cfg:/home/smriprep/.nipype/nipype.cfg \
```
effigies: Does nipype know to look there?

oesteban: That's $HOME within docker and doesn't make any difference to most singularity installations

Dockerfile
```diff
@@ -150,22 +150,6 @@ ENV HOME="/home/smriprep"
...
-               tfapi.get(['MNI152Lin', 'MNI152NLin2009cAsym', 'OASIS30ANTs'], desc='brain', suffix='mask'); \
-               tfapi.get('OASIS30ANTs', resolution=1, desc='4', suffix='dseg'); \
-               tfapi.get(['OASIS30ANTs', 'NKI'], resolution=1, label='brain', suffix='probseg'); \
-               tfapi.get(['MNI152NLin2009cAsym', 'OASIS30ANTs', 'NKI'], resolution=1, \
-                         desc='BrainCerebellumExtraction', suffix='mask'); \
-               tfapi.get(['MNI152NLin2009cAsym', 'OASIS30ANTs', 'NKI'], resolution=1, \
-                         desc='BrainCerebellumRegistration', suffix='mask');" && \
-    rm setup.cfg && \
-    find $HOME/.cache/templateflow -type d -exec chmod go=u {} + && \
-    find $HOME/.cache/templateflow -type f -exec chmod go=u {} +
```
effigies: I'm not sure why we wouldn't want to bundle these in images.

oesteban: I think mounting the templateflow folder from outside is more effective for sMRIPrep. I probably would not propagate this to fMRIPrep.

Pros:
- The image is not rebuilt every time there's a change on the setup.cfg file (any change, not just templateflow's pin).
- The image is a bit lighter and the dockerfile simpler
- It is easier for nipype to use cached results, and we might want to roll it back to timestamp hashing at some point.
- The pybids pin was ineffective because of the installation of templateflow, which already installed 0.9.4 as a dependency. When smriprep was installed, it identified pybids was available regardless of the pin.
- Saves build time in the long run, as templateflow will identify all these files as cached.

Cons:
- Time-cost of this PR, perhaps?

oesteban:
Good to go?

PR #163, opened 2020-01-31
ENH: Fix template keys output in normalization workflow, when cohort present

oesteban:
Getting poldracklab/fmriprep#1955 in a healthy state.

### DIFF
--- a/smriprep/workflows/outputs.py
+++ b/smriprep/workflows/outputs.py
@@ -63,9 +63,6 @@ def init_anat_reports_wf(reportlets_dir, freesurfer,
         DerivativesDataSink(base_directory=reportlets_dir, suffix='T1w'),
         name='ds_std_t1w_report', run_without_submitting=True)
 
-    def _drop_cohort(in_template):
-        return in_template.split(':')[0]
-
     workflow.connect([
         (inputnode, tf_select, [(('template', _drop_cohort), 'template'),
                                 ('template_spec', 'template_spec')]),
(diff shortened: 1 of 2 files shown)


smriprep/workflows/outputs.py
```diff
@@ -64,7 +64,7 @@ def init_anat_reports_wf(reportlets_dir, freesurfer,
         name='ds_std_t1w_report', run_without_submitting=True)
 
     workflow.connect([
-        (inputnode, tf_select, [('template', 'template'),
+        (inputnode, tf_select, [(('template', _drop_cohort), 'template'),
```
mgxd: why not just connect from `split_desc.name` since that is dropping the cohort?

smriprep/workflows/outputs.py
```diff
@@ -193,23 +193,23 @@ def init_anat_derivatives_wf(bids_root, freesurfer, num_t1w, output_dir,
         # Template
         (inputnode, ds_t1w_tpl_warp, [
             ('anat2std_xfm', 'in_file'),
-            ('template', 'to')]),
+            (('template', _drop_cohort), 'to')]),
```
mgxd: same here

smriprep/workflows/outputs.py
```diff
@@ -193,23 +193,23 @@ def init_anat_derivatives_wf(bids_root, freesurfer, num_t1w, output_dir,
...
         (inputnode, ds_t1w_tpl_inv_warp, [
             ('std2anat_xfm', 'in_file'),
-            ('template', 'from')]),
+            (('template', _drop_cohort), 'from')]),
```
mgxd: here

smriprep/workflows/outputs.py
oesteban: name is not exposed outside the workflow - can we live with this inline?

mgxd: this connection is within workflow, no?
i'm suggesting changing it to 
```suggestion
        (split_desc, tf_select, [('name', 'template'),
```
to avoid an excess function call

oesteban: I can't find `split_desc` in this module... :D

PR #167, opened 2020-02-11
ENH: Add option to skip brain extraction

### LINKED ISSUE #12: Gracefully handle pre-skull-stripped T1w images

mgxd:
Supercedes #116 
Fixes #12.

### DIFF
--- a/smriprep/cli/run.py
+++ b/smriprep/cli/run.py
@@ -88,8 +88,9 @@ def get_parser():
                         help='do not use a random seed for skull-stripping - will ensure '
                              'run-to-run replicability when used with --omp-nthreads 1')
     g_ants.add_argument(
-        '--skip-brain-extraction', required=False, action='store_true',
-        default=False, help='Skip the brain extraction workflow (for pre-extracted T1ws))')
+        '--skip-brain-extraction', required=False, action='store',
+        default='auto', help='Skip the brain extraction workflow. '
+                             'Options: auto (default), skip, force. ')
 
     # FreeSurfer options
     g_fs = parser.add_argument_group('Specific options for FreeSurfer preprocessing')
--- a/smriprep/cli/run.py
+++ b/smriprep/cli/run.py
@@ -88,9 +88,10 @@ def get_parser():
                         help='do not use a random seed for skull-stripping - will ensure '
                              'run-to-run replicability when used with --omp-nthreads 1')
     g_ants.add_argument(
-        '--skip-brain-extraction', required=False, action='store',
-        default='auto', help='Skip the brain extraction workflow. '
-                             'Options: auto (default), skip, force. ')
+        '--skull-strip-mode', action='store', choices=('auto', 'skip', 'force'),
+        default='auto', help='determiner for T1-weighted skull stripping (force ensures skull '
+                             'stripping, skip ignores skull stripping, and auto automatically '
+                             'ignores skull stripping if pre-stripped brains are detected).')
 
     # FreeSurfer options
     g_fs = parser.add_argument_group('Specific options for FreeSurfer preprocessing')
@@ -411,8 +412,8 @@ def build_workflow(opts, retval):
         omp_nthreads=omp_nthreads,
         output_dir=str(output_dir),
         run_uuid=run_uuid,
-        skip_brain_extraction=opts.skip_brain_extraction,
         skull_strip_fixed_seed=opts.skull_strip_fixed_seed,
+        skull_strip_mode=opts.skull_strip_mode,
         skull_strip_template=opts.skull_strip_template[0],
         spaces=opts.output_spaces,
         subject_list=subject_list,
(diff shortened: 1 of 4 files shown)


mgxd:
cc @adelavega - I added a few documentation misses, and hopefully avoided missing any merges

smriprep/workflows/anatomical.py
```diff
@@ -68,10 +72,11 @@ def init_anat_preproc_wf(
                 freesurfer=True,
                 hires=True,
                 longitudinal=False,
-                num_t1w=1,
+                t1w=['T1w.nii.gz'],
```
effigies: Maybe add to the run-up to `init_anat_preproc_wf`:

```Python
import nibabel as nb
import numpy as np
nb.Nifti1Image([[[1]]], np.eye(4)).to_filename('T1w.nii.gz')
```

Not sure if you can do any fancy fixture to make sure it's in a temporary directory. Or you can just remove it after.

adelavega:
Looks reasonable to me, thanks!

smriprep/workflows/anatomical.py
mgxd: we might be able to dust off the `wf_pre_code` option https://nipype.readthedocs.io/en/0.13.1/users/sphinx_ext.html#configuration-options, though i'm not entirely certain how to use it / we'll still need to remove the file

smriprep/cli/run.py
```diff
@@ -87,6 +87,10 @@ def get_parser():
     g_ants.add_argument('--skull-strip-fixed-seed', action='store_true',
                         help='do not use a random seed for skull-stripping - will ensure '
                              'run-to-run replicability when used with --omp-nthreads 1')
+    g_ants.add_argument(
+        '--skip-brain-extraction', required=False, action='store',
+        default='auto', help='Skip the brain extraction workflow. '
```
mgxd: the help message should be expanded to reflect on each action's behavior

smriprep/workflows/anatomical.py
```diff
@@ -104,6 +111,10 @@ def init_anat_preproc_wf(
         Enable debugging outputs
     name : :obj:`str`, optional
         Workflow name (default: anat_preproc_wf)
+    skip_brain_extraction : :obj:`str`
+        Skip ants brain extraction workflow, and instead use N4-only workflow
```
mgxd: docstring should also reflect choices

mgxd:
I'm leaning towards including this in 20.1, but if there are strong sentiments we can cut smriprep 0.6 and include this in 20.0

adelavega:
Awesome, thanks. No rush from me at all.

effigies:
Okay. Let's not rush it then. I think we'll have a 0.5.2 release for fMRIPrep 20.0, and we can merge this afterwards.

smriprep/workflows/anatomical.py
```diff
@@ -222,12 +233,34 @@ def init_anat_preproc_wf(
...
+        extreme planes for all (near) zeros.
+        """
+        def _check_img(img):
+            data = np.abs(nb.load(img).get_fdata(dtype=np.float32))
+            sidevals = data[0, :, :].sum() + data[-1, :, :].sum() + \
+                data[:, 0, :].sum() + data[:, -1, :].sum() + \
+                data[:, :, 0].sum() + data[:, :, -1].sum()
+            return sidevals < 10
+
+        return all(_check_img(img) for img in imgs)
```
oesteban: as we are delaying this definition, let's delay it to the latest possible (added suggestion below)
```suggestion
```

smriprep/workflows/anatomical.py
```diff
@@ -222,12 +233,34 @@ def init_anat_preproc_wf(
...
+
+    if skull_strip_mode == 'auto':
+        skull_strip_mode = _is_skull_stripped(t1w)
```
oesteban:
```suggestion
        import numpy as np
        import nibabel as nb
        def _is_skull_stripped(imgs):
            """
            Check if T1w images are skull-stripped.
            
            An input is called skull-stripped if (almost) all voxels in extreme planes are zero.
            
            """
            def _check_img(img):
                data = np.abs(nb.load(img).get_fdata(dtype=np.float32))
                sidevals = data[0, :, :].sum() + data[-1, :, :].sum() + \
                    data[:, 0, :].sum() + data[:, -1, :].sum() + \
                    data[:, :, 0].sum() + data[:, :, -1].sum()
                return sidevals < 10

            return all(_check_img(img) for img in imgs)
        skull_strip_mode = _is_skull_stripped(t1w)
```

smriprep/workflows/anatomical.py
```diff
@@ -3,6 +3,9 @@
 """Anatomical reference preprocessing workflows."""
 from pkg_resources import resource_filename as pkgr
 
+import numpy as np
+import nibabel as nb
+
```
oesteban: Please delay import
```suggestion
```

PR #173, opened 2020-04-02
MAINT: bump dependencies

mgxd:
This bumps the minimum niworkflows version to be in line with fmriprep 20.1.x

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -25,12 +25,12 @@ install_requires =
     matplotlib >= 2.2.0
     nibabel >= 3.0.1
     nipype >= 1.3.1
-    niworkflows ~= 1.1.7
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -25,12 +25,12 @@ install_requires =
     matplotlib >= 2.2.0
     nibabel >= 3.0.1
     nipype >= 1.3.1
-    niworkflows ~= 1.1.7
+    niworkflows >= 1.2.0rc1, < 1.3
     numpy
     packaging
-    pybids >= 0.9.4
+    pybids >= 0.10.2
```
oesteban, line 31: This is not really necessary yet, I believe.

PR #175, opened 2020-04-10
MAINT: pin troublesome sphinx

mgxd:


mgxd:
the troublesome docs build is passing, merging.

Issue #176, opened 2020-04-15
Mishandling of outputs: dseg segmentation derivative different from workflow


oesteban:
In the context of integrating #107 with *fMRIPrep*, I've found that the labels of the `_dseg` volume we are writing out into the derivatives folder are BIDS-compliant, but they are not (as they come from FSL FAST) for the output node of the anatomical workflow.

*fMRIPrep* expects 3 to be the white-matter label (i.e., non-BIDS) as it receives the dseg from the outputnode.

We are beefing up the manipulations that happen within datasinks (or immediately before), so it would make sense for all output nodes to pull their contents out of the datasinks instead of from the workflow. Otherwise, our datasinks should be bare sinks without any further logic.

This particular instance:

https://github.com/poldracklab/smriprep/blob/c0435778fe3145645d33b94e1d0850e545b8dffc/smriprep/workflows/outputs.py#L136-L137

Issue #170, opened 2020-03-08
Replace FSL FAST with FreeSurfer's aseg for recon-all pipeline


oesteban:
For two reasons:
- Economy (no need to run a simpler, extra methodology when a better segmentation is available)
- Run-to-run reproducibility (because FreeSurfer's aseg is less sensitive to isolated pixels that may vary on the brain mask run-to-run, plus the easy reuse of the FS folder)

PR #177, opened 2020-04-15
FIX: Correct the ``dseg`` labeling from FSL FAST earlier

oesteban:
Postponing this relabeling created a deviation of the derivative written
out to the hard disk and the actual file passed on by the outputnode.

This PR should serve as a basis for the elimination of FSL FAST in
``recon-all`` -based workflows (cc/ @mgxd):

  - FAST and the new tailgating node should be moved inside the if
    branch for ``--fs-no-reconall``
  - The LUT mapping function should be reused to remap FS's ``aseg``
    into a 3-tissue brain segmentation.

Resolves: #176
Related: #170
Related: poldracklab/fmriprep#1175

### DIFF
--- a/smriprep/utils/misc.py
+++ b/smriprep/utils/misc.py
@@ -1,0 +1,23 @@
+"""Self-contained utilities to be used within Function nodes."""
+
+
+def apply_lut(in_dseg, lut, newpath=None):
+    """Map the input discrete segmentation to a new label set (lookup table, LUT)."""
+    import numpy as np
+    import nibabel as nb
+    from nipype.utils.filemanip import fname_presuffix
+
+    if newpath is None:
+        from os import getcwd
+        newpath = getcwd()
+
+    out_file = fname_presuffix(in_dseg, suffix='_dseg', newpath=newpath)
+    lut = np.array(lut, dtype='int16')
+
+    segm = nb.load(in_dseg)
+    hdr = segm.header.copy()
+    hdr.set_data_dtype('int16')
+    segm.__class__(lut[np.asanyarray(segm.dataobj, dtype=int)].astype('int16'),
+                   segm.affine, hdr).to_filename(out_file)
+
+    return out_file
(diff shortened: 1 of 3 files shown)


smriprep/utils/misc.py
```diff
@@ -0,0 +1,23 @@
...
+    if newpath is None:
+        from os import getcwd
+        newpath = getcwd()
+
+    out_file = fname_presuffix(in_dseg, suffix='_dseg', newpath=newpath)
+    lut = np.array(lut, dtype='int16')
+
+    segm = nb.load(in_dseg)
+    hdr = segm.header.copy()
+    hdr.set_data_dtype('int16')
```
mgxd, line 19: why is this being copied and altered if you're saving the original header?

oesteban: Good catch! I should be setting this header

oesteban: Mended

oesteban:
Good to go?

smriprep/workflows/anatomical.py
```diff
@@ -320,14 +321,20 @@ def _check_img(img):
...
 
-    # 3. Brain tissue segmentation
+    # 3. Brain tissue segmentation - FAST produces: 0 (bg), 1 (wm), 2 (csf), 3 (gm)
     t1w_dseg = pe.Node(fsl.FAST(segments=True, no_bias=True, probability_maps=True),
                        name='t1w_dseg', mem_gb=3)
 
+    # Change LookUp Table - BIDS wants: 0 (bg), 1 (gm), 2 (wm), 3 (csf)
+    lut_t1w_dseg = pe.Node(niu.Function(function=_apply_bids_lut),
+                           name='lut_t1w_dseg')
+    lut_t1w_dseg.inputs.lut = [0, 3, 1, 2]  # Maps: 0 -> 0, 3 -> 1, 1 -> 2, 2 -> 3.
```
mgxd, line 331: we don't expect this to change so let's just make it immutable
```suggestion
    lut_t1w_dseg.inputs.lut = (0, 3, 1, 2)  # Maps: 0 -> 0, 3 -> 1, 1 -> 2, 2 -> 3.
```

Issue #178, opened 2020-04-17
Inconsistent ordering of tissue maps botches fast-tracked operation


oesteban:
The ordering of tissues here (imposed by the output of FSL FAST, in this case):

https://github.com/poldracklab/smriprep/blob/076bda9f34e27a3d359fb2f2577d0953d57e3cc6/smriprep/workflows/outputs.py#L168

And here:

https://github.com/poldracklab/smriprep/blob/076bda9f34e27a3d359fb2f2577d0953d57e3cc6/smriprep/data/io_spec.json#L20

should be matched, otherwise, the tissues will be wrongly assigned by *fMRIPrep* when operating with #107.

PR #179, opened 2020-04-17
FIX: Make TPMs label ordering in ``io_spec.json`` consistent with workflow

oesteban:
Resolves: #178

### DIFF
--- a/smriprep/data/io_spec.json
+++ b/smriprep/data/io_spec.json
@@ -17,7 +17,7 @@
       },
       "tpms": {
         "datatype": "anat",
-        "label": [ "CSF", "WM", "GM"],
+        "label": [ "CSF", "GM", "WM"],
         "suffix": "probseg"
       }
     },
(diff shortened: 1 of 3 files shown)


oesteban:
If changes of 80a99d3 work (and there's no reason to believe they won't), they would confirm this PR is going the right direction and making things consistent.

oesteban:
Okay, it worked: the colors of contours in the anatomical (first) plot match (because the underlying labels match) - that wasn't happening before.

- DS005: [fasttrack](https://1544-164755300-gh.circle-artifacts.com/0/fasttrack/smriprep/sub-01.html#Anatomical), and [regular](https://1544-164755300-gh.circle-artifacts.com/0/tmp/ds005/derivatives/smriprep/sub-01.html#Anatomical).
- DS054: [fasttrack](https://1545-164755300-gh.circle-artifacts.com/0/fasttrack-derivatives/smriprep/sub-100185.html#Anatomical), and [regular](https://1545-164755300-gh.circle-artifacts.com/0/tmp/ds054/derivatives/smriprep/sub-100185.html#Anatomical).

In the interest of time, I'll go ahead and merge this.

PR #180, opened 2020-04-17
ENH: Execute FSL FAST only with ``--fs-no-reconall``

oesteban:
This PR restructures the anatomical workflow so that the ``recon-all``
based graph does not execute FSL FAST, and rather projects the
``aseg`` segmentation to a 3-tissue brain segmentation with labels
according to BIDS.

This should provide the confounds workflow of *fMRIPrep* with smoother
(and probably more accurate) regions (e.g., for aCompCor calculations).
On the long run, we hope this will add some extra run-to-run
reproducibility to the aCompCor calculations.

Resolves: #170

### DIFF
--- a/smriprep/workflows/anatomical.py
+++ b/smriprep/workflows/anatomical.py
@@ -634,6 +634,18 @@ def _pop(inlist):
 
 
 def _aseg_to_three():
+    """
+    Map FreeSurfer's segmentation onto a brain (3-)tissue segmentation.
+
+    This function generates an index of 255+0 labels and maps them into zero (bg),
+    1 (GM), 2 (WM), or 3 (CSF). The new values are set according to BIDS-Derivatives.
+    Then the index is populated (e.g., label 3 in the original segmentation maps to label
+    1 in the output).
+    The `aseg lookup table
+    <https://github.com/freesurfer/freesurfer/blob/2beb96c6099d96508246c14a24136863124566a3/distribution/ASegStatsLUT.txt>`__
+    is available in the FreeSurfer source.
+
+    """
     import numpy as np
     # Base struct
     aseg_lut = np.zeros((256,), dtype="int")
--- a/smriprep/workflows/anatomical.py
+++ b/smriprep/workflows/anatomical.py
@@ -431,7 +431,7 @@ the brain-extracted T1w using `fast` [FSL {fsl_ver}, RRID:SCR_002823,
         return workflow
 
     # Map FS' aseg labels onto three-tissue segmentation
-    lut_t1w_dseg.inputs.lut = _aseg_two_three()
+    lut_t1w_dseg.inputs.lut = _aseg_to_three()
     split_seg = pe.Node(niu.Function(function=_split_segments),
                         name='split_seg')
 
@@ -633,7 +633,7 @@ def _pop(inlist):
     return inlist
 
 
-def _aseg_two_three():
+def _aseg_to_three():
     import numpy as np
     # Base struct
     aseg_lut = np.zeros((256,), dtype="int")
@@ -676,10 +676,9 @@ def _split_segments(in_file):
     hdr.set_data_dtype('uint8')
 
     out_files = []
-    for i, label in enumerate(("CSF", "WM", "GM")):
-        out_fname = str((Path() / f"aseg_label-{label}_mask.nii.gz").absolute())
-        segment = (data == i + 1).astype('uint8')
-        segimg.__class__(segment, segimg.affine, hdr).to_filename(out_fname)
+    for i, label in enumerate(("CSF", "WM", "GM"), 1):
+        out_fname = str(Path.cwd() / f"aseg_label-{label}_mask.nii.gz")
+        segimg.__class__(data == i, segimg.affine, hdr).to_filename(out_fname)
         out_files.append(out_fname)
 
     return out_files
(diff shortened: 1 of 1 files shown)


smriprep/workflows/anatomical.py
```diff
@@ -620,3 +631,55 @@ def _pop(inlist):
...
+    import nibabel as nb
+
+    segimg = nb.load(in_file)
+    data = np.int16(segimg.dataobj)
+    hdr = segimg.header.copy()
+    hdr.set_data_dtype('uint8')
+
+    out_files = []
+    for i, label in enumerate(("CSF", "WM", "GM")):
+        out_fname = str((Path() / f"aseg_label-{label}_mask.nii.gz").absolute())
```
effigies, line 680: `p.absolute()` means exactly `Path.cwd() / p`, so this is clearer:

```suggestion
        out_fname = str(Path.cwd() / f"aseg_label-{label}_mask.nii.gz")
```

smriprep/workflows/anatomical.py
```diff
@@ -620,3 +631,55 @@ def _pop(inlist):
...
+    import numpy as np
+    import nibabel as nb
+
+    segimg = nb.load(in_file)
+    data = np.int16(segimg.dataobj)
+    hdr = segimg.header.copy()
+    hdr.set_data_dtype('uint8')
+
+    out_files = []
+    for i, label in enumerate(("CSF", "WM", "GM")):
```
effigies, line 679: Clearer than adding 1 later, IMO:

```suggestion
    for i, label in enumerate(("CSF", "WM", "GM"), 1):
```

smriprep/workflows/anatomical.py
```diff
@@ -620,3 +631,55 @@ def _pop(inlist):
...
+        out_fname = str((Path() / f"aseg_label-{label}_mask.nii.gz").absolute())
+        segment = (data == i + 1).astype('uint8')
+        segimg.__class__(segment, segimg.affine, hdr).to_filename(out_fname)
```
effigies, line 682: You've already set the dtype. Since you're not inspecting the image, there's no need to pre-coerce bools to uint8 for consistency. This can be done very briefly:

```suggestion
        segimg.__class__(data == i, segimg.affine, hdr).to_filename(out_fname)
```

smriprep/workflows/anatomical.py
```diff
@@ -620,3 +631,55 @@ def _pop(inlist):
     if isinstance(inlist, (list, tuple)):
         return inlist[0]
     return inlist
+
+
+def _aseg_two_three():
```
effigies, line 636: A link to the reference used to build this would be a helpful comment.

oesteban: There's no link to this. I just scanned the freesurfer LUT and created a mapping myself (I had done something similar for the carpetplot). I tried to find some existing (prescribed with freesurfer tools) to do this task but found nothing. However, there are some labels for FSL-FAST segmentations in the freesurfer's LUT file - that makes me think there must be something.

I have looked at the generated segmentations and they look great - but some particular labels can be arguably changed (e.g., there are two lateral vessels that probably should be mapped to zero or to CSF, but because they are defined in a GM cluster of labels, I mistakenly included them).

You also found a naming problem to this function. My intention was:
```suggestion
def _aseg_to_three():
```

smriprep/workflows/anatomical.py
```diff
@@ -423,13 +413,28 @@ def _check_img(img):
...
+            (buffernode, t1w_dseg, [('t1w_brain', 'in_files')]),
+            (t1w_dseg, lut_t1w_dseg, [('tissue_class_map', 'in_dseg')]),
+            (t1w_dseg, anat_norm_wf, [
+                ('probability_maps', 'inputnode.moving_tpms')]),
+            (t1w_dseg, outputnode, [('probability_maps', 't1w_tpms')]),
         ])
         return workflow
 
+    # Map FS' aseg labels onto three-tissue segmentation
+    lut_t1w_dseg.inputs.lut = _aseg_two_three()
```
oesteban, line 434:
```suggestion
    lut_t1w_dseg.inputs.lut = _aseg_to_three()
```

smriprep/workflows/anatomical.py
effigies: What about https://github.com/freesurfer/freesurfer/blob/release_6_0_0/distribution/ASegStatsLUT.txt?

smriprep/workflows/anatomical.py
```diff
@@ -620,3 +631,54 @@ def _pop(inlist):
     if isinstance(inlist, (list, tuple)):
         return inlist[0]
     return inlist
+
+
+def _aseg_to_three():
```
oesteban, line 636:
```suggestion
def _aseg_to_three():
    """
    Map FreeSurfer's segmentation onto a brain (3-)tissue segmentation.
    
    This function generates an index of 255+0 labels and maps them into zero (bg),
    1 (GM), 2 (WM), or 3 (CSF). The new values are set accordingly to BIDS-Derivatives.
    Then the index is populated (e.g., label 3 in the original segmentation maps to label
    1 in the output).
```

oesteban: @effigies, what about something like this?

effigies: Sure. Though `s/accordingly/according/`, and I might say "The `aseg lookup table <...>`__ is available in the FreeSurfer source.".

PR #183, opened 2020-04-29
ENH: Use new ``DerivativesDataSink`` from NiWorkflows 1.2.0

oesteban:
This PR integrates the new API that will be released with 1.2.0.

References: nipreps/niworkflows#507


oesteban:
Locally checking the correct indexing of `probseg` and `dseg` results, w/wo FreeSurfer. Once that is done, this is all yours, @mgxd.

setup.cfg
```diff
@@ -79,8 +79,6 @@ smriprep =
     data/boilerplate.bib
     data/io_spec.json
     data/itkIdentityTransform.txt
-    data/reports/config.json
```
oesteban, line 82: These are relics of the old reports system.

smriprep/cli/run.py
```diff
@@ -258,7 +257,7 @@ def _warn_redirect(message, category, filename, lineno, file=None, line=None):
 
         logger.log(25, 'Writing reports for participants: %s', ', '.join(subject_list))
         # Generate reports phase
-        errno += generate_reports(subject_list, output_dir, work_dir, run_uuid,
+        errno += generate_reports(subject_list, output_dir, run_uuid,
```
oesteban, line 260: reports are directly written out to `<output_dir>/sub-i/figures/`

smriprep/cli/run.py
```diff
@@ -397,9 +396,8 @@ def build_workflow(opts, retval):
         logger.log(25, 'Running --reports-only on participants %s', ', '.join(subject_list))
         if opts.run_uuid is not None:
             run_uuid = opts.run_uuid
-        retval['return_code'] = generate_reports(
-            subject_list, str(output_dir), str(work_dir), run_uuid,
-            config=pkgrf('smriprep', 'data/reports/config.json'))
```
oesteban, line 402: This was still using the old system. I'm afraid the --reports-only option is not sufficiently tested.

smriprep/workflows/anatomical.py
```diff
@@ -425,8 +421,9 @@ def _check_img(img):
             (buffernode, t1w_dseg, [('t1w_brain', 'in_files')]),
             (t1w_dseg, lut_t1w_dseg, [('tissue_class_map', 'in_dseg')]),
             (t1w_dseg, anat_norm_wf, [
-                ('probability_maps', 'inputnode.moving_tpms')]),
-            (t1w_dseg, outputnode, [('probability_maps', 't1w_tpms')]),
+                (('probability_maps', _probseg_fast2bids), 'inputnode.moving_tpms')]),
```
oesteban, line 424: This reorders the probseg files for bids. It is relevant because this is the easiest way of propagating them, and assigning them the correct `label-` by the DerivativesDataSink.

smriprep/workflows/outputs.py
```diff
@@ -136,15 +135,13 @@ def init_anat_derivatives_wf(bids_root, freesurfer, num_t1w, output_dir,
...
-                            compress=True),
+        DerivativesDataSink(base_directory=output_dir, suffix='dseg', compress=True),
         name='ds_t1w_dseg', run_without_submitting=True)
 
     ds_t1w_tpms = pe.Node(
-        DerivativesDataSink(base_directory=output_dir,
-                            suffix='probseg', compress=True),
+        DerivativesDataSink(base_directory=output_dir, suffix='probseg', compress=True),
         name='ds_t1w_tpms', run_without_submitting=True)
-    ds_t1w_tpms.inputs.extra_values = ['label-CSF', 'label-GM', 'label-WM']
```
oesteban, line 147: bye bye extra_values :)

smriprep/workflows/outputs.py
```diff
@@ -171,17 +168,16 @@ def init_anat_derivatives_wf(bids_root, freesurfer, num_t1w, output_dir,
...
     ds_t1w_tpl_inv_warp = pe.Node(
-        DerivativesDataSink(base_directory=output_dir, allowed_entities=['from', 'to', 'mode'],
-                            to='T1w', mode='image', suffix='xfm'),
+        DerivativesDataSink(base_directory=output_dir, to='T1w', mode='image', suffix='xfm'),
         name='ds_t1w_tpl_inv_warp', run_without_submitting=True)
 
     ds_t1w_tpl_warp = pe.Node(
-        DerivativesDataSink(base_directory=output_dir, allowed_entities=['from', 'to', 'mode'],
-                            mode='image', suffix='xfm', **{'from': 'T1w'}),
+        DerivativesDataSink(base_directory=output_dir, mode='image', suffix='xfm',
```
oesteban, line 179: Outstanding problem (coming from before this PR): how should we encode the cohort if several are used? perhaps with a `+` symbol, if bids-standard/bids-specification#226 (or part of it) finally goes forward?

like

```
sub-01_from-T1w_to-MNIPediatric+4_mode-image_xfm.h5
sub-01_from-T1w_to-MNIPediatric+5_mode-image_xfm.h5
```
where 4, and 5 indicate those cohorts.

smriprep/workflows/outputs.py
```diff
@@ -239,25 +235,25 @@ def init_anat_derivatives_wf(bids_root, freesurfer, num_t1w, output_dir,
...
+        DerivativesDataSink(base_directory=output_dir,
                             mode='image', to='fsnative', suffix='xfm', **{'from': 'T1w'}),
         name='ds_t1w_fsnative', run_without_submitting=True)
     ds_fsnative_t1w = pe.Node(
-        DerivativesDataSink(base_directory=output_dir, allowed_entities=['from', 'to', 'mode'],
+        DerivativesDataSink(base_directory=output_dir,
                             mode='image', to='T1w', suffix='xfm', **{'from': 'fsnative'}),
         name='ds_fsnative_t1w', run_without_submitting=True)
     # Surfaces
-    name_surfs = pe.MapNode(GiftiNameSource(
```
oesteban, line 254: After this, we can deprecate `GiftiNameSource` if we think we won't be using it anymore.

setup.cfg
```diff
@@ -30,7 +30,7 @@ install_requires =
     packaging
     pybids >= 0.10.2
     pyyaml
-    templateflow >= 0.5.2
+    templateflow ~= 0.6
```
mgxd, line 33: it seems we keep flip flopping on `~=` vs `>=`

smriprep/workflows/anatomical.py
```diff
@@ -694,3 +691,7 @@ def _split_segments(in_file):
         out_files.append(out_fname)
 
     return out_files
+
+
+def _probseg_fast2bids(inlist):
```
mgxd, line 696: can you add a comment specifying expected inputs / outputs?

smriprep/workflows/outputs.py
mgxd: i thought that was solved (albeit hacky) before the PR by squishing together `space-MNIPediatric_cohort-1`? did we lose the ability to produce different cohorts / resolutions of the same template?

setup.cfg
oesteban: I would make sure we fulfill poldracklab/fmriprep#2054, and since TemplateFlow is not there there is a little more flexibility.

With the integration of the auto-update, it feels safe to say `~= 0.6`. For the TemplateFlow client there is no API change in the horizon that would affect fMRIPrep. If any is introduced, we should/can start its 1.x series (and therefore the `~=` selector would nicely work out).

smriprep/workflows/outputs.py
oesteban: sure, the `space_<id>_cohort-<n>` is not a problem. The problem comes when you want to write a transform file and there are several cohorts of the same template being used. In that situation, you'll need to cram the cohort field somewhere in the derivative name, and currently there is no way to do that.

smriprep/workflows/anatomical.py
oesteban, line 696:
```suggestion
def _probseg_fast2bids(inlist):
    """Reorder a list of probseg maps from FAST (CSF, WM, GM) to BIDS (GM, WM, CSF)."""
```

PR #188, opened 2020-05-13
FIX: Convert LTA to ITK with nitransforms

### LINKED ISSUE #185: Bringing T1w image into template space
Hello.
I have a multi-session study and am trying to bring a T1w image into template space using `antsApplyTransforms` with 
`sub-{subject}/ses-{session}/anat/*_T1w_space-orig_target-T1w_affine.txt`

However, the resulting image is completely flipped around (orientation labels are off) and cropped. The T1w is LAS oriented, but bringing it to RAS first didn't do anything?

Do you have any advise how to proceed? Thanks.

effigies:
Closes #185.

### DIFF
--- a/smriprep/workflows/anatomical.py
+++ b/smriprep/workflows/anatomical.py
@@ -617,7 +617,7 @@ A T1w-reference map was computed after registration of
         (merge_xfm, concat_xfms, [('out', 'in_xfms')]),
         # Output
         (t1w_reorient, outputnode, [('out_file', 't1w_ref')]),
-        (concat_xfms, outputnode, [('out_inv', 't1w_realign_xfm')]),
+        (concat_xfms, outputnode, [('out_xfm', 't1w_realign_xfm')]),
     ])
 
     return workflow
--- a/smriprep/workflows/anatomical.py
+++ b/smriprep/workflows/anatomical.py
@@ -18,10 +18,10 @@ from nipype.interfaces.ants import N4BiasFieldCorrection
 from niworkflows.engine.workflows import LiterateWorkflow as Workflow
 from niworkflows.interfaces.freesurfer import (
     StructuralReference,
-    PatchedConcatenateLTA as ConcatenateLTA,
     PatchedLTAConvert as LTAConvert,
 )
 from niworkflows.interfaces.images import TemplateDimensions, Conform, ValidateImage
+from niworkflows.interfaces.nitransforms import ConcatenateXFMs
 from niworkflows.interfaces.utility import KeySelect
 from niworkflows.utils.misc import fix_multi_T1w_source_name, add_suffix
 from niworkflows.anat.ants import init_brain_extraction_wf, init_n4_only_wf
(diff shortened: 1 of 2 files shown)


smriprep/interfaces/nitransforms.py
```diff
@@ -0,0 +1,81 @@
...
+    if reference is not None and not xfm.reference:
+        xfm.reference = reference
+
+    xfm.to_filename(out_file, moving=moving, fmt=fmt)
+
+    if out_inv is not None:
+        inv_xfm = ~xfm
+        if moving is not None:
+            inv_xfm.reference = moving
+        (~xfm).to_filename(out_inv, moving=reference, fmt=fmt)
```
oesteban, line 81: I don't know why I did this, but it looks like I really wanted to do this:
```suggestion
        inv_xfm.to_filename(out_inv, moving=reference, fmt=fmt)
```

I spotted this in fMRIPrep but failed to report. Leaving the note here for future consideration.

oesteban:
BTW, would this (=nitransforms integration) be part of 20.1?

effigies:
Hmm. Was that not already party of fmriprep?

oesteban:
Correct.

I integrated nitransforms for that problem because the alternative seemed even finickier (nipype conditionals and more third-party tooling). Nitransforms allowed me to read LTA (when running the full pipeline) and ITK (when using the anatomical fast-track) very easily.

I guess we can say the experimenting with nitransforms is done, and proceed with a more global (or _full_ using my own term) integration.

I'm happy with that, just making sure we are all on the same page 👍

effigies:
For 20.1, I'd say let's transition only things that need adjustment for some reason. We can make a full transition an explicit goal of 20.2.

oesteban:
Okay, how that approach materializes in practice? I mean, you are proposing:

a) having fmriprep's and smriprep's copies of this code coliving until 20.2; or
b) upstreaming the code to niworkflows (so there's no duplicity), but refrain from including more features of nitransforms beyond this interface, until 20.2

effigies:
b, more or less.

I just meant "let's not transition convert3d nodes that are working". If it seems worth it to upstream the interface to niworkflows because it's now in two places, that's okay.

oesteban:
sounds great. I'll send a PR to niworkflows moving the code, tomorrow.

smriprep/workflows/anatomical.py
```diff
@@ -614,12 +612,12 @@ def _set_threads(in_list, maximum):
...
-        (t1w_conform_xfm, concat_affines, [('out_lta', 'in_lta1')]),
-        (t1w_merge, concat_affines, [('transform_outputs', 'in_lta2')]),
-        (concat_affines, lta_to_itk, [('out_file', 'in_lta')]),
+        (t1w_conform_xfm, merge_xfm, [('out_lta', 'in1')]),
+        (t1w_merge, merge_xfm, [('transform_outputs', 'in2')]),
+        (merge_xfm, concat_xfms, [('out', 'in_xfms')]),
         # Output
         (t1w_reorient, outputnode, [('out_file', 't1w_ref')]),
-        (lta_to_itk, outputnode, [('out_itk', 't1w_realign_xfm')]),
+        (concat_xfms, outputnode, [('out_inv', 't1w_realign_xfm')]),
```
oesteban, line 620: I have a vague memory that FreeSurfer's LTAs nomenclature is like the BIDS-Derivatives' `mode-points` of transforms (i.e., the _reference_ is the `to-` field and its data is moved to the _target_ --or `from-`). I'll need to check this offline to make sure we are correctly using and naming these outputs.

oesteban: I haven't seen any ApplyTransform based on these new ITK transforms in sMRIPrep, so I'll need to run a full fMRIPrep run with this to make sure we got the direction of the transform correct.

oesteban:
Running this on ds031 right now. I'll report back tomorrow

oesteban:
Okay, I had to switch from ds031 to ds114 as datalad+openneuro wasn't up-to-date when I started. Now I have sMRIPrep run on ds114 - will report soon whether this is working properly.

oesteban:
Because of #192, I haven't locally tested https://github.com/poldracklab/smriprep/pull/188/commits/a5a826326fa3daa1c53e95f7ae59b0e6c28b51b7 - will report back when finished. This should not affect Circle builds as we don't have any multisession / multi-t1 dataset there.

smriprep/workflows/anatomical.py
oesteban: Addressed in https://github.com/poldracklab/smriprep/pull/188/commits/8db93dcff385c2aabd15f626e706bcaa8239d7d5. Checked on one session+run of ds031.

oesteban:
@effigies, @mgxd any final comments?

(Mathias, we can release right now and add it later. Otherwise, the changelog will need an update - WDYT?)

effigies:
LGTM.

mgxd:
yeah, this will make it in for 20.1

Issue #189, opened 2020-05-19
Two inline functions (with fMRIPrep)


oesteban:
Here https://github.com/poldracklab/smriprep/commit/65e75f31c3e3d81ea14118ac031a3455e7d1bf93#diff-816eb5ff17b071ecf371cf48577d5a3cR426 I introduced this inline function which conflicts with another inline function in fMRIPrep.

```
200519-08:05:29,290 nipype.workflow CRITICAL:
         fMRIPrep failed: Does not support two inline functions in series ('def _probseg_fast2bids(inlist):
    """Reorder a list of probseg maps from FAST (CSF, WM, GM) to BIDS (GM, WM, CSF)."""
    return (inlist[1], inlist[2], inlist[0])
'  and 'def _pick_csf(files):
    return files[0]
'), found when connecting fmriprep_wf.single_subject_100068_wf.anat_preproc_wf.t1w_dseg to fmriprep_wf.single_subject_100068_wf.func_preproc_task_machinegame_run_001_wf.bold_confounds_wf.csf_roi. Please use a Function node.
```

To preempt this from happening again, let's add to the coding standards guide that it is not allowed to connect outputnodes with attached inline functions on the connection.

PR #190, opened 2020-05-19
FIX: Revise tissue probability maps connections and order

oesteban:
This PR amends reminding problems in the label assigments of tissue
probability maps after #183.

Resolves: #189
References: #183

### DIFF
--- a/smriprep/workflows/anatomical.py
+++ b/smriprep/workflows/anatomical.py
@@ -413,6 +413,8 @@ the brain-extracted T1w using `fast` [FSL {fsl_ver}, RRID:SCR_002823,
         t1w_dseg = pe.Node(fsl.FAST(segments=True, no_bias=True, probability_maps=True),
                            name='t1w_dseg', mem_gb=3)
         lut_t1w_dseg.inputs.lut = (0, 3, 1, 2)  # Maps: 0 -> 0, 3 -> 1, 1 -> 2, 2 -> 3.
+        fast2bids = pe.Node(niu.Function(function=_probseg_fast2bids), name="fast2bids",
+                            run_without_submitting=True)
 
         workflow.connect([
             (brain_extraction_wf, buffernode, [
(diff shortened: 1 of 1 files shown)


smriprep/workflows/anatomical.py
```diff
@@ -685,7 +686,7 @@ def _split_segments(in_file):
     hdr.set_data_dtype('uint8')
 
     out_files = []
-    for i, label in enumerate(("CSF", "WM", "GM"), 1):
+    for i, label in enumerate(("GM", "WM", "CSF"), 1):
```
mgxd, line 689: this matches up with bids derivatives 👍 

does `segimg` match up to these labels? if so, we were producing the wrong inputs for some time?

oesteban: Looks like so, which puzzles me a bit because I checked several times on this.

smriprep/workflows/anatomical.py
```diff
@@ -413,17 +413,18 @@ def _check_img(img):
...
             (brain_extraction_wf, buffernode, [
                 (('outputnode.out_file', _pop), 't1w_brain'),
                 ('outputnode.out_mask', 't1w_mask')]),
             (buffernode, t1w_dseg, [('t1w_brain', 'in_files')]),
             (t1w_dseg, lut_t1w_dseg, [('tissue_class_map', 'in_dseg')]),
-            (t1w_dseg, anat_norm_wf, [
-                (('probability_maps', _probseg_fast2bids), 'inputnode.moving_tpms')]),
-            (t1w_dseg, outputnode, [
-                (('probability_maps', _probseg_fast2bids), 't1w_tpms')]),
+            (t1w_dseg, fast2bids, [('tissue_class_map', 'inlist')]),
```
mgxd, line 425:
```suggestion
            (t1w_dseg, fast2bids, [('probability_maps', 'inlist')]),
```

oesteban: Oh my

smriprep/workflows/anatomical.py
oesteban: Please note that this was an incidental finding - I was checking the correctness on both ``--fs-no-reconall`` and with reconall paths and discovered this.

There are two more combinations to this, corresponding to the equivalent paths using the ``--anat-derivatives`` flag.

oesteban: This has been out with rc4 and rc5. Have we released fMRIPrep 20.0.x depending on any of these RCs?

oesteban: Okay, nope (https://github.com/poldracklab/fmriprep/blob/20.0.7/setup.cfg). Which is great news - thanks @effigies for the version dependency matrix - that saved our asses this time :)

oesteban:
After checking on the potential impact of the bug, I think we should go ahead with this one (and cut an rc6).

mgxd:
@oesteban sounds good, though i don't follow

> There are two more combinations to this, corresponding to the equivalent paths using the --anat-derivatives flag.

are there more bugs?

oesteban:
I meant that we need to make sure we are now correctly generating the tissue probability maps under 4 conditions:
- With `recon-all`
- Without `recon-all`
- With fast-track of `recon-all`
- With fast-track, without `recon-all`

mgxd:
ok, looking into this:
✅ with recon-all
✅ without recon-all

i'm unsure if there's any difference with/out fast-track. I'm fine with this getting merged whenever you're ready

oesteban:
The fast-track is being exercised by CircleCI, so I'm less concerned about those paths (and I believe that is the root cause for this bug to exist at all, that we stopped running the full workflow on Circle). Thanks for the testing, let's wait for your runs to finish :+1:

PR #194, opened 2020-05-22
PIN: niworkflows ~= 1.2.0, nipype >= 1.4.0

oesteban:


setup.cfg
```diff
@@ -25,7 +25,7 @@ install_requires =
     matplotlib >= 2.2.0
     nibabel >= 3.0.1
     nipype ~= 1.4
```
mgxd, line 27:
```suggestion
    nipype >= 1.4
```

PR #205, opened 2020-06-07
FIX: Adapt to the new FS canary interface (backwards compatible)

### LINKED ISSUE #203: New FS canary broke master
See #201 broken builds:

https://app.circleci.com/pipelines/github/poldracklab/smriprep/599/workflows/4e52113a-01ac-4d2a-938a-a85d9bb1ad77/jobs/3457/steps
https://app.circleci.com/pipelines/github/poldracklab/smriprep/599/workflows/4e52113a-01ac-4d2a-938a-a85d9bb1ad77/jobs/3455/steps

oesteban:
Addresses the problem discussed in nipreps/niworkflows#538 with a backwards compatible change (i.e., the canary function should work regardless of the nipreps 1.2.x used).

Closes #203.

### DIFF
--- a/smriprep/cli/run.py
+++ b/smriprep/cli/run.py
@@ -176,14 +176,8 @@ def build_opts(opts):
     warnings.showwarning = _warn_redirect
 
     # Precedence: --fs-license-file, $FS_LICENSE, default_license
-    fs_home = os.getenv("FREESURFER_HOME")
-    license_file = (
-        opts.fs_license_file
-        or os.getenv('FS_LICENSE')
-        or (os.path.join(fs_home, "license.txt") if fs_home else None)
-    )
-    if license_file and os.path.isfile(license_file):
-        os.environ["FS_LICENSE"] = license_file
+    if opts.fs_license_file is not None:
+        os.environ["FS_LICENSE"] = opts.fs_license_file
 
     if not check_valid_fs_license():
         raise RuntimeError(
--- a/smriprep/cli/run.py
+++ b/smriprep/cli/run.py
@@ -176,8 +176,16 @@ def build_opts(opts):
     warnings.showwarning = _warn_redirect
 
     # Precedence: --fs-license-file, $FS_LICENSE, default_license
-    license_file = opts.fs_license_file or os.getenv('FS_LICENSE')
-    if not check_valid_fs_license(lic=license_file):
+    fs_home = os.getenv("FREESURFER_HOME")
+    license_file = (
+        opts.fs_license_file
+        or os.getenv('FS_LICENSE')
+        or (os.path.join(fs_home, "license.txt") if fs_home else None)
+    )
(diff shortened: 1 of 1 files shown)


smriprep/cli/run.py
```diff
@@ -176,8 +176,16 @@ def _warn_redirect(message, category, filename, lineno, file=None, line=None):
...
-    license_file = opts.fs_license_file or os.getenv('FS_LICENSE')
-    if not check_valid_fs_license(lic=license_file):
+    fs_home = os.getenv("FREESURFER_HOME")
+    license_file = (
+        opts.fs_license_file
+        or os.getenv('FS_LICENSE')
+        or os.path.join(fs_home, "license.txt") if fs_home else None
+    )
+    if os.path.isfile(license_file):
+        os.environ["FS_LICENSE"] = license_file
```
effigies, line 186: Note that the operator precedence makes this

```Python
(opts.fs_license_file
 or os.getenv('FS_LICENSE')
 or os.path.join(fs_home, "license.txt")
) if fs_home else None
```

I think this is fine, because `fs_home` has to be defined for FreeSurfer to work at all, so `license_file` will always be a string, but if it were None `os.path.isfile` would fail.

But I think the logic should be:

```suggestion
    if opts.fs_license_file is not None:
        os.environ["FS_LICENSE"] = opts.fs_license_file
```

`check_valid_fs_license` will find `$FS_LICENSE` or `$FREESURFER_HOME/license.txt`, so we don't need to modify the environment except in this one case.

oesteban: This is exactly the point, `check_valid_fs_license` needs to be operated assuming it is ignorant of the environment. It needs just to run `mri_convert` and inform out about the result.

The most effective way of making multiproc workers to be able to run FS is just setting FS_LICENSE.

oesteban: Addressed the problem of precedence (i think)

effigies: The code I suggested allows check_valid_fs_license to be ignorant of the environment.

oesteban: Sorry, I misunderstood your comments. I just pushed a commit with your suggestion.

smriprep/cli/run.py
```diff
@@ -176,8 +176,10 @@ def _warn_redirect(message, category, filename, lineno, file=None, line=None):
     warnings.showwarning = _warn_redirect
 
     # Precedence: --fs-license-file, $FS_LICENSE, default_license
-    license_file = opts.fs_license_file or os.getenv('FS_LICENSE')
-    if not check_valid_fs_license(lic=license_file):
+    if opts.fs_license_file is not None:
+        os.environ["FS_LICENSE"] = opts.fs_license_file
```
effigies, line 180: Stringify and make absolute:

```suggestion
        os.environ["FS_LICENSE"] = os.path.abspath(opts.fs_license_file)
```

Issue #41, opened 2017-04-27
Handle IsRunning files in recon-all


effigies:
recon-all creates `IsRunning` files to indicate that it is currently running, and blocks other instances of recon-all from running. Some crashes aren't recovered from gracefully enough for recon-all to remove the file, resulting in a need to manually remove the file.

We have a few options:

1) Leave it to the user to remove; `recon-all` gives a reasonably thorough error message
2) Always remove; assume that nobody will ever run two instances of fmriprep on the same subject
3) Sometimes remove, based on a flag that indicates the user knows what they're doing?

I'm open to suggestions. I'm a little inclined to 1 or 3.

oesteban:
Bumping this up. I think that option 3 + checking the mtime of the `recon-all.log` gives us the best tradeoff. If mtime < 24h then the warning when deleting IsRunning files is noisier and warns about the risks of running competing fmriprep processes.

PR #207, opened 2020-06-08
ENH: Add function to handle stale IsRunning files

mgxd:
This is a proof of concept for what was discussed in #41

A few things to note:
- The only way `IsRunning` files will be removed is if they have a modification time of `> 1 day/24 hours/86400 seconds` from the time of running.
- This is implemented as an overwrite Node, meaning it will always rerun.
- The function always returns the `subject_dir`, which is then passed to `surface_recon_wf` to ensure any necessary cleaning has occured.

I will add some testing if we decide to go this route

### DIFF
--- a/smriprep/utils/misc.py
+++ b/smriprep/utils/misc.py
@@ -55,15 +55,12 @@ def fs_isRunning(subjects_dir, subject_id, mtime_tol=86400):
     isrunning = tuple(subj_dir.glob("**/IsRunning*"))
     if not isrunning:
         return subjects_dir
-    try:
-        reconlog = next(subj_dir.glob("**/recon-all.log"))
-    except StopIteration:
-        # problems finding recon-all.log, just return subjects_dir
-        return subjects_dir
-
-    # if any IsRunning files are found, first check on last modified time of log
-    if (time.time() - reconlog.stat().st_mtime) < mtime_tol:
-        return subjects_dir
+    reconlog = subj_dir / "scripts" / "recon-all.log"
+    if not reconlog.is_file() or (time.time() - reconlog.stat().st_mtime) < mtime_tol:
+        raise RuntimeError(f"{subj_dir} contains IsRunning files: {isrunning}\n"
+                           "FreeSurfer will not run if these are present, to avoid "
+                           "interfering with a running process. If no process is running, "
+                           "they may be safely removed.")
 
     for fl in isrunning:
         fl.unlink()
--- a/smriprep/utils/misc.py
+++ b/smriprep/utils/misc.py
@@ -58,10 +58,12 @@ def fs_isRunning(subjects_dir, subject_id, mtime_tol=86400, logger=None):
     # if recon log doesn't exist, just clear IsRunning
     mtime = reconlog.stat().st_mtime if reconlog.exists() else 0
     if (time.time() - mtime) < mtime_tol:
-        raise RuntimeError(f"{subj_dir} contains IsRunning files: {isrunning}\n"
-                           "FreeSurfer will not run if these are present, to avoid "
-                           "interfering with a running process. If no process is running, "
-                           "they may be safely removed.")
+        raise RuntimeError(f"""\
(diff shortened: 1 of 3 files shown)


smriprep/utils/misc.py
```diff
@@ -21,3 +21,53 @@ def apply_lut(in_dseg, lut, newpath=None):
...
+        return subjects_dir
+    try:
+        reconlog = next(subj_dir.glob("**/recon-all.log"))
+    except StopIteration:
+        # problems finding recon-all.log, just return subjects_dir
+        return subjects_dir
+
+    # if any IsRunning files are found, first check on last modified time of log
+    if (time.time() - reconlog.stat().st_mtime) < mtime_tol:
+        return subjects_dir
```
effigies, line 66: We know at this point that it will not successfully run. Does it make sense instead to raise a readable exception?

```suggestion
    reconlog = subj_dir / "scripts" / "recon-all.log"
    if not reconlog.is_file() or time.time() - reconlog.stat().st_mtime < mtime_tol:
        raise RuntimeError("{subj_dir} contains IsRunning files: {isrunning}\n"
                           "FreeSurfer will not run if these are present, to avoid "
                           "interfering with a running process. If no process is running, "
                           "they may be safely removed.")
```

smriprep/utils/misc.py
```diff
@@ -21,3 +21,53 @@ def apply_lut(in_dseg, lut, newpath=None):
...
+    from pathlib import Path
+    import time
+
+    if subjects_dir is None:
+        return subjects_dir
+    subj_dir = Path(subjects_dir) / subject_id
+    if not subj_dir.exists():
+        return subjects_dir
+
+    isrunning = tuple(subj_dir.glob("**/IsRunning*"))
```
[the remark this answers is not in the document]
mgxd: this is how I originally had it, but wasn't sure how/if BIDS handle freesurfer recons across different sessions?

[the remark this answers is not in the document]
oesteban: Do you mean running with `--longitudinal`?

For the default case, the T1w will be merged in one single-template so freesurfer should not need to handle sessions.

smriprep/utils/misc.py
```diff
@@ -21,3 +21,50 @@ def apply_lut(in_dseg, lut, newpath=None):
...
+    if not isrunning:
+        return subjects_dir
+    reconlog = subj_dir / "scripts" / "recon-all.log"
+    if not reconlog.is_file() or (time.time() - reconlog.stat().st_mtime) < mtime_tol:
```
oesteban, line 59:
```suggestion
    if reconlog.is_file() and (time.time() - reconlog.stat().st_mtime) < mtime_tol:
```
Can we assume that if `reconlog` does not exist then the freesurfer folder is stale and the IsRunning files can safely be removed? (cc @effigies)

mgxd: i've changed the behavior in 76baa26

smriprep/utils/misc.py
```diff
@@ -21,3 +21,49 @@ def apply_lut(in_dseg, lut, newpath=None):
...
+    if not isrunning:
+        return subjects_dir
+    reconlog = subj_dir / "scripts" / "recon-all.log"
+    # if recon log doesn't exist, just clear IsRunning
+    mtime = reconlog.stat().st_mtime if reconlog.exists() else 0
+    if (time.time() - mtime) < mtime_tol:
+        raise RuntimeError(f"{subj_dir} contains IsRunning files: {isrunning}\n"
+                           "FreeSurfer will not run if these are present, to avoid "
+                           "interfering with a running process. If no process is running, "
+                           "they may be safely removed.")
```
oesteban, line 64:
```suggestion
        raise RuntimeError(f"""\
The FreeSurfer's subject folder <{subj_dir}> contains IsRunning files that \
may pertain to a current or past execution: {isrunning}.
FreeSurfer cannot run if these are present, to avoid interfering with a running \
process. Please, make sure no other process is running ``recon-all`` on this subject \
and proceed to delete the files listed above.""")
```

Issue #142, opened 2019-11-26
Ensure workflow docstrings include all inputnode/outputnode slots


mgxd:
For example, https://github.com/poldracklab/smriprep/blob/3213c38fdaf35a2fc7880fd8532d4086d5bd5463/smriprep/workflows/anatomical.py#L123-L161
is missing `template`, `t1w_realign_xfm`, `t1w_aseg`, `t1w_aparc`

https://github.com/poldracklab/smriprep/blob/3213c38fdaf35a2fc7880fd8532d4086d5bd5463/smriprep/workflows/anatomical.py#L200-L208

PR #208, opened 2020-06-08
RF: Make workflows keyword-only (PEP 3102)

mgxd:
This change enforces all `init_*` workflows to be keyword-only. We've generally stuck by the practice anyways when initializing workflows

https://www.python.org/dev/peps/pep-3102/

### DIFF
--- a/smriprep/workflows/norm.py
+++ b/smriprep/workflows/norm.py
@@ -16,6 +16,7 @@ from ..interfaces.templateflow import TemplateFlowSelect, TemplateDesc
 
 
 def init_anat_norm_wf(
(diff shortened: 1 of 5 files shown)


oesteban:
While you're at it, do you think you could also address #142 here?

smriprep/workflows/norm.py
```diff
@@ -99,7 +103,7 @@ def init_anat_norm_wf(
 
     """
     ntpls = len(templates)
-    workflow = Workflow('anat_norm_wf')
```
oesteban, line 102: good catch!

Issue #212, opened 2020-06-13
Forward FAST's PVE maps instead of posteriors


oesteban:
As in the original implementation of aCompCor (ref: poldracklab/fmriprep#2129)

PR #213, opened 2020-06-13
FIX: Progress partial volume maps instead of posteriors (FSL FAST)

oesteban:
Selects the PVE maps and PVE MAP segmentation from FSL FAST to forward it through the workflow. PVE maps are smoother (due to the regularization of the MRF), and they are actually used by aCompCor in its original implementation by Behzadi (see poldracklab/fmriprep#2129).

This PR changes the contents of some derivatives and therefore targets the next minor release.

Resolves: #212.


oesteban:
I've checked locally and this works fine. It looks pretty uncontroversial and merging will help me better address poldracklab/fmriprep#2130, so I'll go ahead and merge.

PR #215, opened 2020-06-16
ENH: Calculate surface Euler Number

### LINKED ISSUE #152: Calculate Euler numbers and place them in the sidecar json of surfaces.

mgxd:
Closes #152


smriprep/workflows/surfaces.py
```diff
@@ -474,6 +482,9 @@ def init_gifti_surface_wf(*, name='gifti_surface_wf'):
         (surface_list, fs2gii, [('out', 'in_file')]),
         (fs2gii, fix_surfs, [('converted', 'in_file')]),
         (inputnode, fix_surfs, [('fsnative2t1w_xfm', 'transform_file')]),
+        (fix_surfs, euler, [('out_file', 'in_file')]),
```
oesteban, line 485: I believe we just need to calculate the Euler number on each hemisphere of the wm surface. We could store both in the metadata of the `_desc-preproc_T1w.nii.gz` image.

.circleci/ds005_outputs.txt
```diff
@@ -19,13 +19,21 @@ smriprep/sub-01/anat/sub-01_dseg.nii.gz
 smriprep/sub-01/anat/sub-01_from-fsnative_to-T1w_mode-image_xfm.txt
 smriprep/sub-01/anat/sub-01_from-T1w_to-fsnative_mode-image_xfm.txt
 smriprep/sub-01/anat/sub-01_hemi-L_inflated.surf.gii
+smriprep/sub-01/anat/sub-01_hemi-L_inflated.surf.json
```
oesteban, line 22: All these contain:

```JSON
{
  "EulerNumber": 2
}
```

effigies: Wrong in what way?

oesteban: AFAIK, the Euler number will always be 2 on surfaces that have been inflated or smoothed, as each hemisphere will be 1.0 (no topological errors).

mgxd: are you saying the FreeSurfer algorithm is wrong? Because the interface just extracts the value printed to console..

effigies: The Euler number is a straightforward value: Vertices + Faces - Edges. This is 2 for anything topologically equivalent to a sphere (no holes, no handles, no pinches), so it's just a way of detecting defects (determining in advance whether the surface can be smoothly deformed to a sphere).

effigies: https://en.wikipedia.org/wiki/Euler_characteristic

oesteban: > are you saying the FreeSurfer algorithm is wrong?

I meant it doesn't make sense to calculate the Euler number in surfaces other than the raw (i.e., no smoothing or inflation) white and the pial. Sorry the comment was very unspecific and confusing.

effigies:
@oesteban Are you now saying that we shouldn't bother calculating the Euler number? It was your suggestion, initially, so I just want to be sure before we close this.

oesteban:
We do want, but only for the white and the pial

effigies:
Okay. If it makes life easier, we can calculate it in pure Python straightforwardly:

```Python
triangles = nb.load(surf_fname).agg_data('triangle')

# Count unique edges
edges = set()
for i, j, k in np.sort(triangles, axis=1):  # Sort to avoid double-counting
    edges |= {(i, j), (i, k), (j, k)}

n_faces = triangles.shape[1]
n_vertices = len(np.unique(triangles))
n_edges = len(edges)
euler = n_vertices - n_edges + n_faces
```

Just thinking ahead to when we might want to add other metadata to surfaces.

.circleci/ds005_outputs.txt
```diff
@@ -21,11 +21,15 @@ smriprep/sub-01/anat/sub-01_from-T1w_to-fsnative_mode-image_xfm.txt
 smriprep/sub-01/anat/sub-01_hemi-L_inflated.surf.gii
 smriprep/sub-01/anat/sub-01_hemi-L_midthickness.surf.gii
 smriprep/sub-01/anat/sub-01_hemi-L_pial.surf.gii
+smriprep/sub-01/anat/sub-01_hemi-L_pial.surf.json
```
effigies, line 24: The extension is `.surf.gii`.

```suggestion
smriprep/sub-01/anat/sub-01_hemi-L_pial.json
```

mgxd: this is going to require a change within DDS, since:
https://github.com/nipreps/niworkflows/blob/dbd2a972d51b171c3914ee08248bfaf5543a757c/niworkflows/interfaces/bids.py#L625-L630

effigies:
Is it worth the effort of coordinating with niworkflows to get this in a 20.2 release? If not, let's defer this.

mgxd:
Since we are planning on revising derivatives to be BIDS-spec compliant, I think so. I'll throw together a quick fix for this.

mgxd:
alright, tests are passing. this has stayed open longer than expected, any lingering concerns?

effigies:
Sorry, can you open a PR on niworkflows to make it easy to inspect the changes? But otherwise, this seems fine.

One thing I'm becoming slightly nervous about while dropping double-extensions like `.surf.json` and `.dtseries.json` is that `_bold.json` can refer to `_bold.nii.gz`, `_bold.func.gii` and `_bold.dtseries.nii`. Will spaces save us, or are we setting ourselves up for pain? If we do need these double-extensions, we should probably appeal to the spec to permit them.

smriprep/workflows/surfaces.py
```diff
@@ -555,3 +567,14 @@ def _check_cw256(in_files):
...
         return ['-noskullstrip', '-cw256']
     return '-noskullstrip'
+
+
+def _surfs_meta(surface_file, euler_number):
+    """Creates metadata dictionary for pial, white matter surfaces"""
+    from pathlib import Path
+
+    surface = Path(surface_file).stem
+    if 'smoothwm' in surface or 'pial' in surface:
```
oesteban, line 577: Sorry for the late realization. Now I understand part of the problem - again, calculating the Euler number on the smoothwm surface will not be really effective. And we don't generate conversions of the wm without smoothing.

Why don't we attach the Euler number into the `desc-preproc_T1w.json` file, which is always present?

```
{
  "EulerNumbers": {"WM": 1.7, "Pial": 1.9}
}
```

oesteban: ha! - I just had said this ... https://github.com/nipreps/smriprep/pull/215/files#r446595593

smriprep/workflows/surfaces.py
```diff
@@ -555,3 +567,14 @@ def _check_cw256(in_files):
     if np.any(fov > 256):
         return ['-noskullstrip', '-cw256']
     return '-noskullstrip'
+
+
+def _surfs_meta(surface_file, euler_number):
```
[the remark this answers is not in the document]
mgxd: sure

smriprep/workflows/surfaces.py
effigies: > calculating the Euler number on the smoothwm surface will not be really effective

Why not? Smoothwm does not change the topology. All surfaces will have an Euler number of 2.

mgxd: I'm confused from your example, the EulerNumber can never be a decimal. 

And, at least from my understanding, the euler number is mainly used a QC check of a surface file - shouldn't this then be tied to the surface file rather than the volumetric data?

oesteban: > I'm confused from your example, the EulerNumber can never be a decimal.

Yup, that's right. I have updated the example with valid values (and also the fact that we have two hemispheres)

> Smoothwm does not change the topology

I believe smoothwm not only includes the smoothing, it also goes through a step to resolve defects. I don't think Rosen et al. used the Euler number from the smoothwm, but rather the wm.

> the euler number is mainly used a QC check of a surface file - shouldn't this then be tied to the surface file rather than the volumetric data?

Correct, but the recent interest in it comes for the fact that it is a good proxy for the image quality of the original T1w image. When I requested this feature, this was my intent. 

That said, there's a little trick in your question: the suffix `_T1w` does not signal whether it is surface or volumetric data (that is hinted/stated by the extension), it signals the measurement type (i.e., physical property). Because these surfaces are reconstructed from the T1w, it would be fair to say the surfaces and corresponding metadata are derived from the T1w (and hence, inherit the suffix).

I think we don't have anything explicit in the BIDS-Derivatives spec when a particular derivative has "two or more parents" (e.g., surface reconstruction with T1w, T2w and FLAIR). I believe using the T1w metadata would be the easier place to find the info, and still compliant with the spec.

effigies:
My impression of this issue/PR was that it was cheap metadata to add to files, but I'm concerned over how much time is going into bike-shedding something that should never be anything other than 2, and it makes me think we have different ideas of the possibilities and stakes.

It's probably worth referencing [ReconAllTableStableV6.0](https://surfer.nmr.mgh.harvard.edu/fswiki/ReconAllTableStableV6.0) and [recon-all](https://github.com/freesurfer/freesurfer/blob/v6.0.0/scripts/recon-all), rather than stating our beliefs of what happens.

`mri_tessellate` creates the `?h.orig.nofix` surfaces, which may have Euler numbers ≠ 2. `mris_fix_topology` creates `?h.orig`, which should have Euler number = 2. If that fails, then `mris_topo_fixer` [is used](https://github.com/freesurfer/freesurfer/blob/2beb96c6099d96508246c14a24136863124566a3/scripts/recon-all#L3337-L3398). Again, the likely result is Euler number = 2.

So our options at this point are that `?h.orig` either has an Euler number of 2, or it doesn't. If it does, then all derived surfaces, including `?h.white`, `?h.smoothwm` and `?h.pial` will be topologically identical, and have the same Euler number. If it doesn't, they will still have the same Euler number, if they can be made. I don't think they can, in which case `recon-all` will fail and fMRIPrep will not complete without errors. But perhaps we should check with the mailing list.

mgxd:
I'm just going to close this implementation, as it ends up having a farther reach than #215 intended anyways.

PR #218, opened 2020-07-28
RF: Structural outputs

mgxd:
Currently, there is a fundamental difference between how smriprep and fmriprep are producing outputs. smriprep outputs will always be in a templateflow-supported resolution, whereas fmriprep allows having its outputs match the native BOLD resolution.

Related issue: https://github.com/poldracklab/fmriprep/issues/2226

### DIFF
--- a/smriprep/workflows/anatomical.py
+++ b/smriprep/workflows/anatomical.py
@@ -325,7 +325,9 @@ the brain-extracted T1w using `fast` [FSL {fsl_ver}, RRID:SCR_002823,
     anat_norm_wf = init_anat_norm_wf(
         debug=debug,
         omp_nthreads=omp_nthreads,
-        templates=spaces.get_spaces(nonstandard=False, dim=(3,)),
+        templates=[
+            str(ref) for ref in spaces.get_standard(full_spec=True, dim=(3,))
+        ],
     )
 
     workflow.connect([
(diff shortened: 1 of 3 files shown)


mgxd:
The latest commit (871093d) submits all specifications to `anat_norm_wf`. (fmriprep compatibility depends on https://github.com/mgxd/fmriprep/commit/bc1a5fde2f488579eb12364b3e0b81c3ea6f2ce7) 

I'm not convince on this method though, as it seems we could have redundant registrations in the case of `<template>:res-1:res-2`. The best bet may be to pass along the desired resolutions, and turn the `tpl_moving, std_mask, std_dseg, std_tpms` Nodes into MapNodes, though I'm unsure how that will affect things downstream.

oesteban:
Very tricky indeed. Here:

https://github.com/nipreps/smriprep/blob/ad75bc9ec534324508afdf11e152d9407dc7c066/smriprep/workflows/outputs.py#L282-L285

the node is parameterized by the `anat2std_xfm` input, which is an iterable. While this is an appealing solution, the workflow is totally unaware of the `--output-spaces`, so passing in any will have no effect.

oesteban:
mmm, I can see now that the conversions in standard spaces are already generated within the normalization workflow - that's why you took that route. I think the best idea would be to strip that off of the normalization.

mgxd:
@oesteban The more I think on this, I feel the normalization workflow should essentially only output the transform(s). That way, users can get all their outputs (T1w normalized, tpms, aparc, etc) in whichever resolution is desired from their `--output-spaces` call. I'll take another pass this coming week if you don't get to it today.

oesteban:
> I feel the normalization workflow should essentially only output the transform(s)

Totally agree.


> I'll take another pass this coming week if you don't get to it today.

I think #219 is pretty close. It is indeed working with ds005 on my desktop. CircleCI will fail, at least for the check of outputs which I haven't updated.

PR #219, opened 2020-08-09
FIX: Generate anatomical conversions with full spec from ``--output-spaces``

oesteban:
Generates the anatomical derivatives in standardized volumetric spaces
directly within the derivatives workflow, instead of taking them as
inputs.

The spaces object must be passed in to determine the full specification
of output spaces.

Has some node duplicity for the generation of the reportlet, but I think
this implementation is much more clear, and follows fMRIPrep's patterns.

Resolves: poldracklab/fmriprep#2226.
Supersedes: #218.

### DIFF
--- a/smriprep/cli/run.py
+++ b/smriprep/cli/run.py
@@ -403,7 +403,8 @@ def build_workflow(opts, retval):
         return retval
 
     output_spaces = opts.output_spaces
-    output_spaces.checkpoint()
+    if not output_spaces.is_cached():
+        output_spaces.checkpoint()
 
     logger.log(25, INIT_MSG(
         version=__version__,
--- a/smriprep/cli/run.py
+++ b/smriprep/cli/run.py
@@ -402,12 +402,15 @@ def build_workflow(opts, retval):
                                                  packagename="smriprep")
         return retval
 
+    output_spaces = opts.output_spaces
+    output_spaces.checkpoint()
+
     logger.log(25, INIT_MSG(
         version=__version__,
         bids_dir=bids_dir,
         subject_list=subject_list,
         uuid=run_uuid,
-        spaces=opts.output_spaces)
+        spaces=output_spaces)
     )
 
     # Build main workflow
@@ -426,7 +429,7 @@ def build_workflow(opts, retval):
         skull_strip_fixed_seed=opts.skull_strip_fixed_seed,
         skull_strip_mode=opts.skull_strip_mode,
         skull_strip_template=opts.skull_strip_template[0],
-        spaces=opts.output_spaces,
+        spaces=output_spaces,
         subject_list=subject_list,
         work_dir=str(work_dir),
         bids_filters=bids_filters,
--- a/smriprep/interfaces/templateflow.py
+++ b/smriprep/interfaces/templateflow.py
@@ -11,7 +11,8 @@ from nipype.interfaces.base import (
 class _TemplateFlowSelectInputSpec(BaseInterfaceInputSpec):
     template = traits.Str('MNI152NLin2009cAsym', mandatory=True, desc='Template ID')
     atlas = InputMultiObject(traits.Str, desc='Specify an atlas')
-    cohort = InputMultiObject(traits.Int, desc='Specify an atlas')
+    cohort = InputMultiObject(traits.Either(traits.Str, traits.Int),
+                              desc='Specify a cohort')
     resolution = InputMultiObject(traits.Int, desc='Specify a template resolution index')
     template_spec = traits.DictStrAny(
         {'atlas': None, 'cohort': None},
--- a/smriprep/interfaces/templateflow.py
+++ b/smriprep/interfaces/templateflow.py
@@ -11,9 +11,13 @@ from nipype.interfaces.base import (
 class _TemplateFlowSelectInputSpec(BaseInterfaceInputSpec):
     template = traits.Str('MNI152NLin2009cAsym', mandatory=True, desc='Template ID')
     atlas = InputMultiObject(traits.Str, desc='Specify an atlas')
+    cohort = InputMultiObject(traits.Int, desc='Specify an atlas')
     resolution = InputMultiObject(traits.Int, desc='Specify a template resolution index')
     template_spec = traits.DictStrAny(
-        {'atlas': None}, usedefault=True, desc='Template specifications')
+        {'atlas': None, 'cohort': None},
+        usedefault=True,
+        desc='Template specifications'
+    )
 
 
 class _TemplateFlowSelectOutputSpec(TraitedSpec):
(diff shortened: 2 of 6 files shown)


oesteban:
This is ready for you @mgxd.

I've checked the following (for ds054 only, because ds005 does not have any normalization in these settings):

- The outputs in standard spaces, when no `res` or `res-native` are given, have the resolution of the original T1w (not that of the template anymore).
- (currently running on circle while writing, but working locally) The outputs in standard spaces can have several specifications (e.g., `MNI152NLin2009cAsym:res-2:res-native`) and that generates the expected conversions (one on the T1w grid and one on res-2 of the corresponding template).

Other comments:
- I have left untouched the normalization workflow, to ensure this PR is minimally invasive. That means conversions of the T1w with the template resolution are generated in the working directory, and they may be generated again if the user sets the corresponding specification (which is `res-1`, except for `--sloppy` runs). These internal conversions are not dumped into the derivatives folder. We may consider minimizing the duplicity (although I don't think it is worth the time).
- I needed to `spaces.checkpoint()` as we are doing in fMRIPrep, to make sure we have a table of references cached.

Please let me know what you think.

oesteban:
BTW, should we target `maint/0.6.x` instead of master?

smriprep/workflows/anatomical.py
```diff
@@ -389,27 +389,24 @@ def _check_img(img):
...
         # Connect derivatives
         (anat_template_wf, anat_derivatives_wf, [
             ('outputnode.t1w_valid_list', 'inputnode.source_files')]),
         (anat_norm_wf, anat_derivatives_wf, [
-            ('poutputnode.template', 'inputnode.template'),
-            ('poutputnode.anat2std_xfm', 'inputnode.anat2std_xfm'),
-            ('poutputnode.std2anat_xfm', 'inputnode.std2anat_xfm')
+            ('outputnode.template', 'inputnode.template'),
+            ('outputnode.anat2std_xfm', 'inputnode.anat2std_xfm'),
+            ('outputnode.std2anat_xfm', 'inputnode.std2anat_xfm')
```
oesteban, line 402: We don't pass the parameterized node anymore, but the JoinNode instead.

smriprep/interfaces/templateflow.py
```diff
@@ -11,9 +11,13 @@
 class _TemplateFlowSelectInputSpec(BaseInterfaceInputSpec):
     template = traits.Str('MNI152NLin2009cAsym', mandatory=True, desc='Template ID')
     atlas = InputMultiObject(traits.Str, desc='Specify an atlas')
+    cohort = InputMultiObject(traits.Int, desc='Specify an atlas')
```
oesteban, line 14: I needed to update this bit. I haven't actually tested this with cohorts though.

smriprep/cli/run.py
```diff
@@ -402,12 +402,15 @@ def build_workflow(opts, retval):
                                                  packagename="smriprep")
         return retval
 
+    output_spaces = opts.output_spaces
+    output_spaces.checkpoint()
```
oesteban, line 406: I needed to checkpoint to be able to read the references' full specs down the line.

smriprep/workflows/outputs.py
```diff
@@ -141,8 +141,16 @@ def init_anat_reports_wf(*, freesurfer, output_dir,
...
 
-def init_anat_derivatives_wf(*, bids_root, freesurfer, num_t1w, output_dir,
-                             name='anat_derivatives_wf', tpm_labels=BIDS_TISSUE_ORDER):
+def init_anat_derivatives_wf(
+    *,
+    bids_root,
+    freesurfer,
+    num_t1w,
+    output_dir,
+    spaces,
```
oesteban, line 150: new argument

smriprep/workflows/outputs.py
```diff
@@ -296,33 +267,33 @@ def init_anat_derivatives_wf(*, bids_root, freesurfer, num_t1w, output_dir,
...
-        (t1w_name, ds_t1w_tpl_warp, [('out', 'source_file')]),
-        (t1w_name, ds_t1w_tpl_inv_warp, [('out', 'source_file')]),
-        (t1w_name, ds_t1w_tpl, [('out', 'source_file')]),
-        (t1w_name, ds_std_mask, [('out', 'source_file')]),
-        (t1w_name, ds_std_dseg, [('out', 'source_file')]),
-        (t1w_name, ds_std_tpms, [('out', 'source_file')]),
     ])
 
+    # Transforms
+    if spaces.get_spaces(nonstandard=False, dim=(3,)):
```
oesteban, line 273: this branch checks whether any standard space was selected at all. If not, there won't be any transforms to write out here.

smriprep/workflows/outputs.py
```diff
@@ -336,6 +307,115 @@ def init_anat_derivatives_wf(*, bids_root, freesurfer, num_t1w, output_dir,
                                           ('t1w_ref_xfms', 'in_file')]),
         ])
 
+    # Write derivatives in standard spaces specified by --output-spaces
+    if getattr(spaces, '_cached') is not None and spaces.cached.references:
```
oesteban, line 311: replicate fmriprep's trick to access this info

smriprep/workflows/outputs.py
```diff
@@ -422,7 +496,9 @@ def _rpt_masks(mask_file, before, after, after_mask=None):
 
 
 def _drop_cohort(in_template):
-    return in_template.split(':')[0]
+    if isinstance(in_template, str):
+        return in_template.split(':')[0]
+    return [_drop_cohort(v) for v in in_template]
```
oesteban, line 501: now this is a list of transforms (not using iterables anymore, but using iterfields). because it is a list, I had to extend the implementation.

smriprep/workflows/outputs.py
```diff
@@ -440,3 +516,20 @@ def _empty_report(in_file=None):
...
+
+
+def _no_native(value):
+    try:
+        return int(value)
+    except Exception:
+        return 1
+
+
+def _drop_path(in_path):
```
oesteban, line 532: I haven't really looked into this - it may not be working (although I would expect an error to be raised if the relative path can't be computed)

oesteban:
Running this locally:
```
smriprep /data/ds000005 derivatives/ participant --participant-label 01 -vv --nthreads 8 --omp-nthreads 8 -w work --fs-subjects-dir /data/ds000005/derivatives/freesurfer-6.0.1/ --output-spaces MNI152NLin2009cAsym MNI152NLin6Asym:res-2:res-native MNIPediatricAsym:cohort-6:res-native:res-2 MNIPediatricAsym:cohort-5 fsaverage:den-10k fsaverage6 --sloppy
```

Will report back on the cohorts handling whenever it finishes.

oesteban:
Okay, this still needs some love to work out the cohorts

oesteban:
The latest commit is working as expected with the command line above. This is good to go on my end, @mgxd

PR #223, opened 2020-08-11
ENH: Retain session info when multi-session data are not averaged

oesteban:
Multi-session data used to be combined, but after adopting ``--bids-filters``, now it is possible to select a single session.

This PR uses a tiny hack, just using the ``num_t1w`` variable already present in the derivatives workflow (it has been added to the reports workflow).

The idea is that session is dropped when ``num_t1w`` > 1.
This works because single-session datasets do not carry any ``session`` information on to the ``DerivativesDataSink``.
So, if ``num_t1w`` is 1 in a multisession dataset, the T1w reference name should carry (and keep) the session entity.

Resolves: #197.

### DIFF
--- a/smriprep/workflows/anatomical.py
+++ b/smriprep/workflows/anatomical.py
@@ -204,6 +204,7 @@ BIDS dataset.""".format(num_t1w=num_t1w)
     # Connect reportlets workflows
     anat_reports_wf = init_anat_reports_wf(
         freesurfer=freesurfer,
+        num_t1w=num_t1w,
         output_dir=output_dir,
     )
     workflow.connect([
--- a/smriprep/workflows/outputs.py
+++ b/smriprep/workflows/outputs.py
@@ -11,7 +11,7 @@ from ..interfaces import DerivativesDataSink
 BIDS_TISSUE_ORDER = ("GM", "WM", "CSF")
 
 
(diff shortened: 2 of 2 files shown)


oesteban:
> Fleeting thought - how does this handle multiple T1w runs from the same session? Is `run` dropped?

Good point - No, this will infer that those T1w are coming from different sessions. Do we want to address that in this PR or file a new issue for the edge case?

I fear that the solution to that edge case will not be as trivial as this PR.

mgxd:
With the near endless combinations available with `--bids-file-filters`, I fear there will be a lot of edge cases. We can add that in a separate PR, this can go ahead.

Looks like merging #219 introduced some conflicts.

effigies:
It looks like we're tackling this problem twice, in `DerivativesDataSink` and `fix_multi_T1w_source_name`. What if we just set `source_file` to be `InputMultiObject`, and extract the common entities from that?

The basic logic should be:

```Python
def common_source(in_files):
    entities = [parse_entities(in_file) for in_file in filename_to_list(in_files)]
    common_entities = {k: v for k, v in entities[0].items()
                       if all(ent.get(k) == v for ent in entities[1:])}
    return build_path(entities)
```

This feels like it would be less brittle, let us deduplicate, and cover @mgxd's use case.

PR #225, opened 2020-08-11
ENH: Retain session info when multi-session data are not averaged

### LINKED ISSUE #197: Retain session information for data that does not combine multiple sessions
If a dataset uses sessions but only one session gets used, we remove it from anatomical and reporting outputs. This has worked because multi-session data was always combined, but now that we have poldracklab/fmriprep#1770, users can request that only a single session be processed.

Given this, the ideal case would be for `ses-<label>` not to be removed from entities unless the derivative combines across sessions. Similarly, if all fMRIPrep-processed data comes from a single session, including it in the report `sub-<label>_ses-<label>.html` will allow for multiple sessions to be processed independently and stored in the same output dataset without conflict.

See [Neurostars #6341](https://neurostars.org/t/anat-vs-func-workflow-treatment-of-sessions/6341).

This would need to be implemented in a new feature release.

effigies:
Stealing a title from #223. Testing nipreps/niworkflows#547.

Closes #197.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -23,7 +23,7 @@ install_requires =
     matplotlib >= 2.2.0
     nibabel >= 3.0.1
     nipype >= 1.4
-    niworkflows ~= 1.2.6
+    niworkflows @ git+https://github.com/effigies/niworkflows@rf/multisource_derivatives
     numpy
     packaging
     pybids >= 0.10.2
--- a/setup.cfg
+++ b/setup.cfg
@@ -23,7 +23,7 @@ install_requires =
     matplotlib >= 2.2.0
     nibabel >= 3.0.1
     nipype >= 1.4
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -23,7 +23,7 @@ install_requires =
     matplotlib >= 2.2.0
     nibabel >= 3.0.1
     nipype >= 1.4
-    niworkflows ~= 1.2.6
+    niworkflows @ git+https://github.com/effigies/niworkflows@rf/multisource_derivatives
```
effigies, line 26:
```suggestion
    niworkflows @ git+https://github.com/nipreps/niworkflows@master

oesteban:
@mgxd any comments before merging in?

mgxd:
nope, let's get it in

PR #233, opened 2020-11-09
ENH: Use 4+ empty edge planes to detect skull-stripped images

effigies:
This proposes a relaxed heuristic for detecting skull-stripped images, detecting each empty edge plane, rather than all edge planes at once. If four planes are empty, we consider it skull-stripped.

This accounts for acquisitions where the brain is cut off on one or two edges.

An alternative we could look at is the actual edges, rather than edge planes. These should be pretty reliably 0 if the image is skullstripped.

Ref: https://github.com/nipreps/smriprep/issues/232#issuecomment-723329317


effigies:
False postives:

* https://openneuro.org/datasets/ds000005/file-display/sub-06:anat:sub-06_T1w.nii.gz

PR #239, opened 2021-02-08
MAINT: Update to new API of *NiWorkflows*

oesteban:
See nipreps/niworkflows#603.


mgxd (commented):
Looks like there still an outdated pointer causing the docs to fail. And the ds005 error will probably be fixed by bumping up numpy in the Docker image to ~1.19.

Otherwise, LGTM

oesteban:
Citeproc is failing but everything else looks in order. I'm going to merge and address the problem separately.

PR #240, opened 2021-02-09
MAINT: Refactor Dockerfile and move tests from travis to GHA

oesteban:


Dockerfile
```diff
@@ -147,6 +141,15 @@ RUN useradd -m -s /bin/bash -G users smriprep
 WORKDIR /home/smriprep
 ENV HOME="/home/smriprep"
 
+# Installing SVGO
+RUN npm install -g svgo
+
+# Installing bids-validator
+RUN npm install -g bids-validator@1.4.0
```
mgxd, line 148: why not group these together and avoid an extra step?

PR #249, opened 2021-06-02
PIN: nipreps/niworkflows bugfix branch

effigies:
Testing https://github.com/nipreps/niworkflows/pull/641.


oesteban:
@mgxd can I ask you to do the same you did for fMRIPrep? (meaning, updating the CI so packaging is only checked on tags and release branches) and update the niworkflows pin to upstream/master?

Once that is done, can this probe PR be closed?

effigies:
This can be closed at will.

PR #257, opened 2021-08-06
FIX: Account for label entity when querying brain mask

mgxd:
UNCInfant template currently fails, since the brain mask contains `label-brain` instead of `desc-brain`. AFAIK, `label` is the preferred entity for this.

This is a quick fix, but looking through the code, I think the `TemplateFlowSelect` interface as a whole needs a substantial rework. Was it envisioned as a nipype wrapper for `templateflow.api.get`?

### DIFF
--- a/smriprep/interfaces/templateflow.py
+++ b/smriprep/interfaces/templateflow.py
@@ -86,6 +86,9 @@ class TemplateFlowSelect(SimpleInterface):
     >>> result.outputs.t1w_file  # doctest: +ELLIPSIS
     '.../tpl-MNIPediatricAsym_cohort-5_res-1_T1w.nii.gz'
 
+    >>> select = TemplateFlowSelect()
+    >>> select.inputs.template = 'UNCInfant:cohort-1'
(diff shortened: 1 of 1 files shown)


oesteban:
> Was it envisioned as a nipype wrapper for `templateflow.api.get`?

Yes

smriprep/interfaces/templateflow.py
```diff
@@ -86,6 +86,9 @@ class TemplateFlowSelect(SimpleInterface):
     >>> result.outputs.t1w_file  # doctest: +ELLIPSIS
     '.../tpl-MNIPediatricAsym_cohort-5_res-1_T1w.nii.gz'
 
+    >>> select = TemplateFlowSelect()
+    >>> select.inputs.template = 'UNCInfant:cohort-1'
+    >>> result = select.run()
```
oesteban, line 91:
```suggestion
    >>> result = select.run()

```

oesteban (approved):

PR #258, opened 2021-08-09
MAINT: CircleCI housekeeping

mgxd:
Following what was implemented in fmriprep:

- Add Docker orb for easy authentication
- Bump machine images to latest LTS (focal)
- Remove setfacl calls

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,7 +32,10 @@ docs_deploy: &docs
         name: Deploy docs to gh-pages branch
         command: gh-pages --dotfiles --message "doc(update) [skip ci]" --dist docs/_build/html
 
-version: 2
+version: 2.1
+orbs:
+  docker: circleci/docker@1.6.0
+
 jobs:
   build:
     machine:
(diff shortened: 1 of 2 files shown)


mgxd:
I'm planning on following fMRIPrep's implementation and switching over to RTD - currently the build is hitting a memory issue (https://readthedocs.org/projects/smriprep/builds/14435765/), I've reached out to RTD support about expanding the build resources.

effigies:
Our docs are set up to be on GH pages, not RTD.

mgxd:
> It re-builds all previous releases each time.

What is the advantage of this? It seems to me RTD would significantly reduce the amount of code we'd have to maintain

effigies:
I think Oscar has mentioned a kind of perennial threat of RTD limiting the free builds in some way.

The rebuild every time is a bug, btw.

mgxd:
Okay, I'm still not 100% convinced (maybe @oesteban can chime in) but for the sake of getting a release out I'll proceed with the current implementation

PR #270, opened 2021-11-05
CI: Drop test for direct invocation of setup.py

effigies:
Direct invocations of setup.py are discouraged by everybody including setuptools developers. The contemporary equivalents of these are pip installing, so let's just test those.


effigies:
Doctest failures seem unrelated. Pushing directly.

PR #272, opened 2021-11-26
[WIP] FIX: Anatomic fast track (except freesurfer/surface rec.)

nuKs:
What my patch does is to rely on pybids layout instead of bids build_path in order to retrieve the fast_track derivatives files.
Pybids layout does adapt to any bids-valid naming scheme, while build_path does not, but rather directly assumes the file name.

So far, examples that make fast-track fails include:

- subjects with a single T1 (as the path is different since no template is made)
- subjects with T1 containing run-x segments within the name.

EXAMPLE:
In my case, IBC has been failing (without fix) for some subjects, modified/reduced day2day datasets too.
IBC openneuro: https://openneuro.org/datasets/ds002685/versions/1.3.1
day2day upon request: https://bmcneurosci.biomedcentral.com/articles/10.1186/s12868-017-0383-y

WARNING:
This fix is less restrictive than previous version, thus might select too much files.
This risk should properly be assessed.

WARNING:
freesurfer files aren't fixed.

cf. https://github.com/nipreps/smriprep/issues/224

### DIFF
--- a/smriprep/utils/bids.py
+++ b/smriprep/utils/bids.py
@@ -26,6 +26,7 @@ from pathlib import Path
 from json import loads
 from pkg_resources import resource_filename as pkgrf
 from bids.layout.writing import build_path
+from bids import BIDSLayout
 
 
 def get_outputnode_spec():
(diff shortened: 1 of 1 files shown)


nuKs:
@oesteban where can I download ds054 (it's not here: https://openfmri.org/dataset/) ?

effigies:
If this still needs doing, let's start over from the current master.

PR #273, opened 2021-12-01
ENH: Register T1w template to FreeSurfer nu.mgz instead of T1.mgz

effigies:
Here's one step toward supporting FastSurfer, which doesn't generate a T1.mgz file.

Via https://github.com/Deep-MI/FastSurfer/issues/21#issuecomment-984049197

Related https://github.com/nipreps/smriprep/issues/226


mgxd:
@effigies should we get this with `0.8.3`?

effigies:
Since FastSurfer support wasn't in 21.0, I would suggest we move this to 0.9.x.

mgxd:
sounds good

effigies:
Based on https://neurostars.org/t/fsnative2t1w-xfm-incorrect-registration/25739/2, it might be better to use `rawavg.mgz`, which it seems like fastsurfer also leaves.

PR #291, opened 2022-07-06
FIX: Disable -T2pial and -FLAIRpial at -autorecon1 stage

effigies:
FreeSurfer 7 will enable T2pial/FLAIRpial processing when passing `-T2` or `-FLAIR` images. This breaks our model of passing in T2w/FLAIR images at -autorecon1, as -T2pial comes much later.

This change moves `-noskullstrip` out of a kind of hidden function and puts it in the workflow, and adds `-noT2pial` and `-noFLAIRpial` to `-autorecon1` to mitigate this bug. This should have no impact on FreeSurfer 6, as recon-all is very forgiving of `-no` flags that do not change anything.


mgxd (approved):
LGTM

Issue #289, opened 2022-06-24
Surface misalignment with non-conformed freesurfer data limits the application of mri_robust_register to oblique native inputs


ericfeczko:
**Describe the bug**
Bug report is extended from nibabies here https://github.com/nipreps/nibabies/issues/225 

I need to make a pull request on smriprep, since this is where its more appropriate. 

**Exact command line executed**
Refer to the above report, which is based off of https://github.com/nipreps/nibabies/issues/222

**Are you positive that the input dataset is BIDS-compliant?**
- [X] I have used the online BIDS-Validator
- [X] I have run a local installation of the BIDS-Validator (please, note the version of the validator here).
- [X] I let *sMRIPrep* check it for me (in other words, **I didn't set the `--skip-bids-validation` argument**).
- [ ] No, I haven't checked myself AND used the `--skip-bids-validation` argument.

***sMRIPrep* feedback information**
Please attach the full log written to the standard output and the crashfile(s), if generated.

**Visual reports of *sMRIPrep***
If applicable, add screenshots or attach the corresponding "reportlets" to help explain your problem.

**Installation type (please complete the following information):**
 - [ ] "Bare-metal"
 - [X] Singularity
 - [ ] Docker

**Additional context**
This is using nibabies 22.1.0.cr1

PR #292, opened 2022-07-06
modified gifti_surface_wf to remove the fsnative transform and replace using to_scanner

ericfeczko:
this pull request address issue #289 -- enabling smriprep to be used with expected inputs for nibabies. Should also improve alignments in edge cases for non-infant processing :)

### DIFF
--- a/smriprep/workflows/surfaces.py
+++ b/smriprep/workflows/surfaces.py
@@ -269,8 +269,6 @@ gray-matter of Mindboggle [RRID:SCR_002438, @mindboggle].
         # reoriented image
         (inputnode, fsnative2t1w_xfm, [('t1w', 'target_file')]),
         (autorecon1, fsnative2t1w_xfm, [('T1', 'source_file')]),
-        (fsnative2t1w_xfm, gifti_surface_wf, [
-            ('out_reg_file', 'inputnode.fsnative2t1w_xfm')]),
         (fsnative2t1w_xfm, t1w2fsnative_xfm, [('out_reg_file', 'in_lta')]),
         # Refine ANTs mask, deriving new mask from FS' aseg
         (inputnode, refine, [('corrected_t1', 'in_anat'),
@@ -531,9 +529,8 @@ def init_gifti_surface_wf(*, name="gifti_surface_wf"):
         run_without_submitting=True,
     )
     fs2gii = pe.MapNode(
-        fs.MRIsConvert(out_datatype="gii"), iterfield="in_file", name="fs2gii"
+        fs.MRIsConvert(out_datatype="gii",to_scanner=True), iterfield="in_file", name="fs2gii"
     )
-    fix_surfs = pe.MapNode(NormalizeSurf(), iterfield="in_file", name="fix_surfs")
 
     # fmt:off
     workflow.connect([
@@ -551,9 +548,7 @@ def init_gifti_surface_wf(*, name="gifti_surface_wf"):
                                       ('inflated', 'in3')]),
         (save_midthickness, surface_list, [('out_file', 'in4')]),
         (surface_list, fs2gii, [('out', 'in_file')]),
-        (fs2gii, fix_surfs, [('converted', 'in_file')]),
-        (inputnode, fix_surfs, [('fsnative2t1w_xfm', 'transform_file')]),
-        (fix_surfs, outputnode, [('out_file', 'surfaces')]),
+        (fs2gii, outputnode, [('converted', 'surfaces')])
     ])
     # fmt:on
     return workflow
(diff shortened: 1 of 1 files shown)


smriprep/workflows/surfaces.py
```diff
@@ -531,9 +529,21 @@ def init_gifti_surface_wf(*, name="gifti_surface_wf"):
...
+    surfmorph_list = pe.Node(
+        niu.Merge(3,ravel_inputs=True),
+        name="surfmorph_list",
+        run_without_submitting=True,
+    )
+    fscurv2funcgii = pe.MapNode(
+        fs.MRIsConvert(out_datatype="gii",to_scanner=True,scalarcurv_file=True),iterfield="in_file", name="fscurv2funcgii"
+    )
+    allsurf_list = pe.Node(
+        niu.merge(7,ravel_inputs=True),
```
mgxd, line 543: is this a list of lists? if not, ravel_inputs isn't needed
```suggestion
        niu.Merge(7),
```

smriprep/workflows/surfaces.py
```diff
@@ -531,9 +529,21 @@ def init_gifti_surface_wf(*, name="gifti_surface_wf"):
...
     fs2gii = pe.MapNode(
-        fs.MRIsConvert(out_datatype="gii"), iterfield="in_file", name="fs2gii"
+        fs.MRIsConvert(out_datatype="gii",to_scanner=True), iterfield="in_file", name="fs2gii"
+    )
+    surfmorph_list = pe.Node(
+        niu.Merge(3,ravel_inputs=True),
+        name="surfmorph_list",
+        run_without_submitting=True,
+    )
+    fscurv2funcgii = pe.MapNode(
```
mgxd, line 539: i'm confused about this step - we aren't doing any functional processing here

smriprep/workflows/surfaces.py
```diff
@@ -531,9 +529,21 @@ def init_gifti_surface_wf(*, name="gifti_surface_wf"):
...
+        fs.MRIsConvert(out_datatype="gii",to_scanner=True,scalarcurv_file=True),iterfield="in_file", name="fscurv2funcgii"
```
mgxd, line 540: the `scalarcurv_file` arg requires an existing file

mgxd (commented):
Thanks for the PR @ericfeczko - can you describe the proposed changes a little bit more in depth? I am a little bit confused.

I went through and quickly left some notes - one thing this would greatly benefit from is running a linter and making sure it passes the flake8 checks.

smriprep/workflows/surfaces.py
ericfeczko: Whoops. This is for outputting the metric surfaces that will be needed for the "goodvoxels" projection fix to eliminate aliasing in the projected outputs.

effigies:
Okay, so I'm reasonably convinced that this is something that needs doing, but we're going to need to continue to use our surface normalization utility, modified to zero out the CRAS metadata without applying it to the coordinates. Here are screenshots of four surfaces viewed in freeview and wbview:

* Red: `mris_convert`
* Green: `mris_convert --to-scanner`
* Blue: `mris_convert` + `NormalizeSurfs` (i.e., add CRAS to points and set CRAS tozero)
* Cyan: `mris_convert --to-scanner` + zero-out CRAS manually in nibabel

###### freeview
freeview
###### wbview
wbview

It's a little hard to see, but basically Blue and Cyan appear correct in both, while red appears correct for freeview and green appears correct for wbview.

I've looked at blue and green, and there is more difference than just the CRAS values (differences can be up to 0.02mm, well above rounding error). So there does seem to be some effect of the obliquity, but I don't understand the transforms involved well enough to reproduce the effect of `--to-scanner` myself.

----

Anyway, my recommendation is to update `NormalizeSurf` to simply zero out the CRAS (we also want to continue adding appropriate metadata to . To keep things isolated, can we separate out the thickness/curv/sulc into a separate PR?

effigies:
Good news: Cyan and green still line up in wbview for infants. Bad news: Nothing lines up for freeview.

infant-freeview
infant-wbview

So we can go ahead with `--to-scanner` + drop CRAS to preserve workbench/FreeSurfer compatibility in reasonably-oriented adults. It's going to take more effort to understand FreeSurfer's interpretation of the two affines to produce GIFTIs that align correctly in both cases.

effigies:
Ah ha! Setting the `VolGeom{X,Y,Z}_{R,A,S}` fields to be `[[-1, 0, 0], [0, 0, 1], [0, -1, 0]]` fixes the cyan to match up with the red in FreeView without damaging it in wbview:

infant-freeview

Issue #288, opened 2022-06-21
adding surface based morphometry outputs (cortical thickness, sulcal depth, folding curvature) to CIFTI-output


ericfeczko:
**Is your feature request related to a problem? Please describe.**
Much of the neuroimaging community from infant neuroimaging to adults wants to measure surface based morphometrics and use cortical thickness/folding curvature/sulcal depth. However, the CIFTI options do not allow for the output of these metrics into derivatives.

**Describe the solution you'd like**
I would like cortical thickness/folding curvature/sulcal depth derived from the freesurfer component to be carried over to the final derivatives.

**Describe alternatives you've considered**
None.

**Additional context**
This is additionally critical for ensuring proper projection of BOLD data to cortical surfaces -- and would help eliminate aliasing issues identified by Guarav's group.

tsalo:
Did #325 address this?

effigies:
Yes.

PR #296, opened 2022-07-21
ENH: Output thickness, curvature, and sulcal depth files

ericfeczko:
This address issue #288. I'm running into some errors with the outputs due to some syntax issues that were introduced. Will be doing some testing with circle CI today to debug. Will also ensure pep8 compatibility.

### DIFF
--- a/smriprep/workflows/surfaces.py
+++ b/smriprep/workflows/surfaces.py
@@ -269,8 +269,6 @@ gray-matter of Mindboggle [RRID:SCR_002438, @mindboggle].
         # reoriented image
         (inputnode, fsnative2t1w_xfm, [('t1w', 'target_file')]),
         (autorecon1, fsnative2t1w_xfm, [('T1', 'source_file')]),
-        (fsnative2t1w_xfm, gifti_surface_wf, [
-            ('out_reg_file', 'inputnode.fsnative2t1w_xfm')]),
         (fsnative2t1w_xfm, t1w2fsnative_xfm, [('out_reg_file', 'in_lta')]),
         # Refine ANTs mask, deriving new mask from FS' aseg
         (inputnode, refine, [('corrected_t1', 'in_anat'),
@@ -531,9 +529,8 @@ def init_gifti_surface_wf(*, name="gifti_surface_wf"):
         run_without_submitting=True,
     )
     fs2gii = pe.MapNode(
-        fs.MRIsConvert(out_datatype="gii"), iterfield="in_file", name="fs2gii"
+        fs.MRIsConvert(out_datatype="gii",to_scanner=True), iterfield="in_file", name="fs2gii"
     )
-    fix_surfs = pe.MapNode(NormalizeSurf(), iterfield="in_file", name="fix_surfs")
 
     # fmt:off
     workflow.connect([
@@ -551,9 +548,7 @@ def init_gifti_surface_wf(*, name="gifti_surface_wf"):
                                       ('inflated', 'in3')]),
         (save_midthickness, surface_list, [('out_file', 'in4')]),
         (surface_list, fs2gii, [('out', 'in_file')]),
-        (fs2gii, fix_surfs, [('converted', 'in_file')]),
-        (inputnode, fix_surfs, [('fsnative2t1w_xfm', 'transform_file')]),
-        (fix_surfs, outputnode, [('out_file', 'surfaces')]),
+        (fs2gii, outputnode, [('converted', 'surfaces')])
     ])
     # fmt:on
     return workflow
--- a/smriprep/workflows/surfaces.py
+++ b/smriprep/workflows/surfaces.py
@@ -535,6 +535,11 @@ def init_gifti_surface_wf(*, name="gifti_surface_wf"):
         name="surfmorph_list",
         run_without_submitting=True,
     )
+    whitesurf_list = pe.Node(
+        niu.Merge(3,ravel_inputs=True),
+        name="whitesurf_list",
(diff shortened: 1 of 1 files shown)


smriprep/workflows/surfaces.py
```diff
@@ -552,9 +567,17 @@ def init_gifti_surface_wf(*, name="gifti_surface_wf"):
...
+                                        ('sulc', 'in2'),
+                                        ('curv', 'in3')]),
+        (get_surfaces, whitesurf_list, [('smoothwm', 'in1'),
+                                        ('smoothwm', 'in2'),
+                                        ('smoothwm', 'in3')]),
+        (whitesurf_list, fscurv2funcgii, [('out', 'in_file')]),
+        (surfmorph_list, fscurv2funcgii, [('out', 'scalarcurv_file')]),
+        (fs2gii, allsurf_list, [('converted', 'in1')]),
+        (fscurv2funcgii, allsurf_list, [('converted', 'in2')]),
+        (allsurf_list, outputnode, [('out', 'surfaces')])
```
effigies, line 580: I think we're going to have trouble conflating `.surf.gii` files and `.shape.gii` files. I think we should probably not combine them and instead make a new `morphometrics` output. Something like:

```suggestion
        (fscurv2funcgii, outputnode, [('converted', 'morphometrics')]),
```

smriprep/workflows/surfaces.py
```diff
@@ -534,7 +531,25 @@ def init_gifti_surface_wf(*, name="gifti_surface_wf"):
...
+    )
+    fscurv2funcgii = pe.MapNode(
+        fs.MRIsConvert(out_datatype="gii", to_scanner=True),
+        iterfield=["in_file", "scalarcurv_file"], name="fscurv2funcgii",
+    )
+    allsurf_list = pe.Node(
+        niu.Merge(7),
+        name="allsurf_list",
+        run_without_submitting=True,
+    )
```
effigies, line 552:
```suggestion
```

smriprep/workflows/surfaces.py
```diff
@@ -534,7 +531,25 @@ def init_gifti_surface_wf(*, name="gifti_surface_wf"):
...
+    whitesurf_list = pe.Node(
+        niu.Merge(3, ravel_inputs=True),
+        name="whitesurf_list",
+        run_without_submitting=True,
+    )
+    fscurv2funcgii = pe.MapNode(
+        fs.MRIsConvert(out_datatype="gii", to_scanner=True),
+        iterfield=["in_file", "scalarcurv_file"], name="fscurv2funcgii",
+    )
```
effigies, line 547: We don't need to iterate if `in_file` is always the same.

```suggestion
    fscurv2funcgii = pe.MapNode(
        fs.MRIsConvert(out_datatype="gii"),
        iterfield=["scalarcurv_file"], name="fscurv2funcgii",
    )
```

smriprep/workflows/surfaces.py
```diff
@@ -552,9 +567,17 @@ def init_gifti_surface_wf(*, name="gifti_surface_wf"):
...
-        (fs2gii, fix_surfs, [('converted', 'in_file')]),
-        (inputnode, fix_surfs, [('fsnative2t1w_xfm', 'transform_file')]),
-        (fix_surfs, outputnode, [('out_file', 'surfaces')]),
+        (get_surfaces, surfmorph_list, [('thickness', 'in1'),
+                                        ('sulc', 'in2'),
+                                        ('curv', 'in3')]),
+        (get_surfaces, whitesurf_list, [('smoothwm', 'in1'),
+                                        ('smoothwm', 'in2'),
+                                        ('smoothwm', 'in3')]),
+        (whitesurf_list, fscurv2funcgii, [('out', 'in_file')]),
```
effigies, line 576:
```suggestion
        (get_surfaces, fscurv2funcgii, [('smoothwm', 'in_file')]),
```

effigies (commented):
If we restore all of the removed lines, we'll preserve the work done in #295. Then I think we can reduce this to a relatively small addition. We'll then need to go into `outputs.py` and add a couple nodes to write out the files correctly, mirroring what we do for `.surf.gii`:

https://github.com/nipreps/smriprep/blob/b29578ec45e51da63e5df6b47a8c88b02ae43acd/smriprep/workflows/outputs.py#L592-L601

smriprep/workflows/surfaces.py
ericfeczko: won't the in_file vary by hemisphere? I'm a little confused by the interface, I'll admit :)

smriprep/workflows/surfaces.py
ericfeczko: that makes a lot of sense -- I'll make sure to do the same moving forward when handling the cifti conversion

smriprep/workflows/surfaces.py
effigies: Oh right. That might be the best way to do this after all.

ericfeczko:
that sounds perfect -- I'm in meeting hell right now, but I can try to give this a shot tonight or tomorrow.

effigies:
Here's an (untested) attempt: 857e0bd8370d967691dc11d9ca7b705e63efb394

You may need to add my fork to try it out:

```
git remote add effigies https://github.com/effigies/smriprep.git
git fetch effigies
git cherry-pick 857e0bd
```

effigies:
What are the expected outputs? I guess `.dscalar.nii` or `.shape.gii`?

mgxd:
Closed by #305

PR #298, opened 2022-08-18
FIX: Update removed `-autorecon-hemi` flag

mgxd:
Fixes https://github.com/nipreps/fmriprep/issues/2832

The `<no>label-exvivo-ec` is now removed, this substitutes it with  `<no>label-exvivo-ec-avg`, though I am not even sure this is necessary anymore?

FS changes from 2015... https://github.com/freesurfer/freesurfer/commit/939c3cbab4f81ef0b54dd2599a591f88dd528ac9

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -738,6 +738,9 @@ jobs:
           name: Install deps
           command: pip install -r docs/requirements.txt
       - run:
+          name: Install package in editable mode
+          command: pip install -e .
+      - run:
           name: Build only this commit
           no_output_timeout: 30m
           command: make -C docs NO_ET=1 SPHINXOPTS="-W" BUILDDIR="_build/no_version_html" html
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -738,8 +738,8 @@ jobs:
           name: Install deps
           command: pip install -r docs/requirements.txt
       - run:
-          name: Install package in editable mode
-          command: pip install -e .
+          name: Install package
+          command: pip install .
       - run:
           name: Build only this commit
           no_output_timeout: 30m
--- a/setup.cfg
+++ b/setup.cfg
@@ -23,6 +23,7 @@ install_requires =
     matplotlib >= 2.2.0
     nibabel >= 4.0.1
     nipype >= 1.7.0
+    traits < 6.4
     niworkflows ~= 1.6.0
     numpy
     packaging
--- a/smriprep/interfaces/freesurfer.py
+++ b/smriprep/interfaces/freesurfer.py
@@ -99,7 +99,7 @@ class _ReconAllInputSpec(fs.preprocess.ReconAllInputSpec):
             "parcstats2",
             "cortparc3",
             "parcstats3",
-            "label-exvivo-ec",
+            "label-exvivo-ec-avg",
             # autorecon vol
             "cortribbon",
             "hyporelabel",  # 6.0
--- a/smriprep/interfaces/freesurfer.py
+++ b/smriprep/interfaces/freesurfer.py
@@ -99,7 +99,8 @@ class _ReconAllInputSpec(fs.preprocess.ReconAllInputSpec):
             "parcstats2",
             "cortparc3",
             "parcstats3",
-            "label-exvivo-ec-avg",
+            "label-exvivo-ec",  # 5.3
+            "label-exvivo-ec-avg",  # 6.0+
             # autorecon vol
             "cortribbon",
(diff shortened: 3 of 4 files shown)


smriprep/interfaces/freesurfer.py
```diff
@@ -99,7 +99,7 @@ class _ReconAllInputSpec(fs.preprocess.ReconAllInputSpec):
             "parcstats2",
             "cortparc3",
             "parcstats3",
-            "label-exvivo-ec",
+            "label-exvivo-ec-avg",
```
effigies, line 102: We have some 5.3 and 6.0 tags. If this ever gets merged back into nipype, it would be good to not throw away 5.3 support needlessly.

```suggestion
            "label-exvivo-ec",  # 5.3
            "label-exvivo-ec-avg",  # 6.0+
```

.circleci/config.yml
```diff
@@ -737,6 +737,9 @@ jobs:
       - run:
           name: Install deps
           command: pip install -r docs/requirements.txt
+      - run:
+          name: Install package in editable mode
+          command: pip install -e .
```
effigies, line 742: Why is this needed?

effigies: It's fine, btw... I'm just skeptical of installing anything in editable mode in CI unless it's to test that editable mode works.

smriprep/interfaces/freesurfer.py
mgxd: will this ignore using the `-<no>label-exvivo-ec` flag on FS6+? I did not realize the commented sections were still being used...

.circleci/config.yml
mgxd: It isn't - the docs/requirement.txt was a mess of niworkflows / documentation dependencies, so cleaned it up

mgxd, line 742:
```suggestion
          name: Install package
          command: pip install .
```

smriprep/interfaces/freesurfer.py
effigies: This just lists the valid steps. I'm not sure where `-<no>label-exvivo-ec` is being used. Could you point to the problem that's happening somewhere?

setup.cfg
```diff
@@ -23,6 +23,7 @@ install_requires =
     matplotlib >= 2.2.0
     nibabel >= 4.0.1
     nipype >= 1.7.0
+    traits < 6.4
```
mgxd, line 26: @effigies should we just cut a new nipype release?

smriprep/interfaces/freesurfer.py
mgxd: So the problem is that `parcstats` is producing the command line:
```
-autorecon-hemi lh -nohyporelabel -lh-only -openmp 8 -subjid sub-XXXXXX -sd /out/sourcedata/freesurfer \
-notessellate -nosmooth1 -noinflate1 -noqsphere -nofix -nowhite -nosmooth2 -noinflate2 -nocurvstats \
-nosphere -nosurfreg -nojacobian_white -noavgcurv -nocortparc -nopial -nopctsurfcon -noparcstats \
-nocortparc2 -noparcstats2 -noparcstats3 -nolabel-exvivo-ec
```

and the `-<no>label-exvivo-ec` flag is no longer valid, causing an instant crash.

setup.cfg
effigies: Yes, I think so.

smriprep/interfaces/freesurfer.py
effigies: Are you able to reproduce this in a debug session? I can't see why it would be happening, since `label-exvivo-ec` is only used [here](https://github.com/nipy/nipype/blob/4a91e88777465b435935d4dcaceff8a8345ba2dd/nipype/interfaces/freesurfer/preprocess.py#L1201), which is guarded by `if Info.looseversion() < LooseVersion("6.0.0"):`.

mgxd: Ah, I am able to reproduce this for nibabies, and see the problem - essentially `Info.looseversion() == 0.0.0` (since the developmental build-stamp.txt is not being correctly parsed), which causes this flag to be included when forcing `recon-all`.

I do not know why this is happening with fmriprep (https://github.com/nipreps/fmriprep/issues/2832) - we'll need to collect / run the `command.txt` and debug that.

mgxd: I believe so, but I have been using a prebuilt distribution instead of building FS (+ the infant module) from source.

Issue #300, opened 2022-09-30
Fit/apply breakdown


effigies:
Ran sMRIPrep on one subject, checking off the proposed things to include in the minimal outputs:

- [ ] sub-02_desc-aparcaseg_dseg.nii.gz
- [ ] sub-02_desc-aseg_dseg.nii.gz
- [x] sub-02_desc-brain_mask.json
- [x] sub-02_desc-brain_mask.nii.gz
- [x] sub-02_desc-preproc_T1w.json
- [x] sub-02_desc-preproc_T1w.nii.gz
- [x] sub-02_dseg.nii.gz
- [x] sub-02_from-fsnative_to-T1w_mode-image_xfm.txt
- [x] sub-02_from-MNI152NLin2009cAsym_to-T1w_mode-image_xfm.h5
- [x] sub-02_from-T1w_to-fsnative_mode-image_xfm.txt
- [x] sub-02_from-T1w_to-MNI152NLin2009cAsym_mode-image_xfm.h5
- [x] sub-02_hemi-L_inflated.surf.gii
- [x] sub-02_hemi-L_midthickness.surf.gii
- [x] sub-02_hemi-L_pial.surf.gii
- [x] sub-02_hemi-L_smoothwm.surf.gii
- [x] sub-02_hemi-R_inflated.surf.gii
- [x] sub-02_hemi-R_midthickness.surf.gii
- [x] sub-02_hemi-R_pial.surf.gii
- [x] sub-02_hemi-R_smoothwm.surf.gii
- [x] sub-02_label-CSF_probseg.nii.gz
- [x] sub-02_label-GM_probseg.nii.gz
- [x] sub-02_label-WM_probseg.nii.gz
- [ ] sub-02_space-MNI152NLin2009cAsym_desc-brain_mask.json
- [ ] sub-02_space-MNI152NLin2009cAsym_desc-brain_mask.nii.gz
- [ ] sub-02_space-MNI152NLin2009cAsym_desc-preproc_T1w.json
- [ ] sub-02_space-MNI152NLin2009cAsym_desc-preproc_T1w.nii.gz
- [ ] sub-02_space-MNI152NLin2009cAsym_dseg.nii.gz
- [ ] sub-02_space-MNI152NLin2009cAsym_label-CSF_probseg.nii.gz
- [ ] sub-02_space-MNI152NLin2009cAsym_label-GM_probseg.nii.gz
[...]

PR #303, opened 2022-09-30
RF: Split anatomical workflow into fit and derivatives workflows

effigies:
Starting on #300.


docs/conf.py
```diff
@@ -65,7 +65,6 @@
     "seaborn",
     "skimage",
     "svgutils",
-    "templateflow",
```
effigies, line 68: Hopefully we don't end up downloading things during builds, but if so, so be it.

smriprep/utils/bids.py
```diff
@@ -50,81 +51,6 @@ def get_outputnode_spec():
     return fields
 
 
-def predict_derivatives(subject_id, output_spaces, freesurfer):
```
effigies, line 53: This didn't seem to be used anywhere, and deleting this freed me up to change `collect_derivatives` to use `BIDSLayout` and use lists as "or" instead of "and" in `io_spec.json`.

smriprep/workflows/anatomical.py
```diff
@@ -36,54 +36,56 @@
...
+    t1w: list,
+    skull_strip_mode: str,
+    skull_strip_template: Reference,
+    spaces: SpatialReferences,
+    precomputed: dict,
+    debug: bool,
+    omp_nthreads: int,
     name="anat_preproc_wf",
-    skull_strip_fixed_seed=False,
+    skull_strip_fixed_seed: bool = False,
```
effigies, line 82: Type annotations grew out of frustration of not knowing what I was working with (particularly Reference and SpatialReference). The consequence of this was that I need to import `Reference` and `SpatialReference` at the module level, which meant that a mocked `templateflow` prevented the loading of spatial references.

smriprep/workflows/apply/resampling.py
```diff
@@ -0,0 +1,151 @@
...
+"""Spatial resampling workflows"""
+from nipype.pipeline import engine as pe
+from nipype.interfaces import utility as niu
+
+from niworkflows.engine.workflows import LiterateWorkflow as Workflow
+from niworkflows.interfaces.fixes import FixHeaderApplyTransforms as ApplyTransforms
+from ...interfaces.templateflow import TemplateFlowSelect
+
+
+def init_resample_template_wf(
```
effigies, line 32: This did not end up getting used. Will probably just drop it. The `workflow.outputs.init_iterator_wf` did everything here that's useful apart from the resampling node.

smriprep/workflows/outputs.py
```diff
@@ -73,137 +77,600 @@ def init_anat_reports_wf(*, freesurfer, output_dir, name="anat_reports_wf"):
...
+                                       ('subject_id', 'subject_id')]),
+            (recon_report, ds_recon_report, [('out_report', 'in_file')]),
+            (inputnode, ds_recon_report, [('source_file', 'source_file')])
+        ])
+        # fmt:on
+
+    return workflow
+
+
+def init_ds_template_wf(
```
effigies, line 235: I feel like we need a better name, but this is the subject template, not MNI or other template spaces.

smriprep/workflows/outputs.py
```diff
@@ -73,137 +77,600 @@ def init_anat_reports_wf(*, freesurfer, output_dir, name="anat_reports_wf"):
...
-            (recon_report, ds_recon_report, [('out_report', 'in_file')]),
-            (inputnode, ds_recon_report, [('source_file', 'source_file')])
-        ])
-        # fmt:on
-
     return workflow
 
 
-def init_anat_derivatives_wf(
+def init_anat_second_derivatives_wf(
```
effigies, line 667: This could definitely stand refactoring, but I'd prefer to keep that until later.

smriprep/workflows/outputs.py
```diff
@@ -655,12 +932,104 @@ def init_anat_derivatives_wf(
     return workflow
 
 
+def init_template_iterator_wf(*, spaces, name="template_iterator_wf"):
```
effigies, line 935: I like this *a lot*. I feel like what we're going to want one of these sitting in the subject level that then gets passed as inputs to the baseline derivatives (registrations only), reporting and full derivative workflows.

Need to spend some time to work out what that would look like, in a separate PR. For now we spin one of these up inside the reporting and full derivatives workflows.

smriprep/workflows/surfaces.py
```diff
@@ -268,36 +234,131 @@ def init_surface_recon_wf(*, omp_nthreads, hires, name="surface_recon_wf"):
...
+            RobustRegister(auto_sens=True, est_int_scale=True), name="fsnative2t1w_xfm"
+        )
+
+        # fmt:off
+        workflow.connect([
+            (inputnode, fsnative2t1w_xfm, [('t1w', 'target_file')]),
+            (autorecon1, fsnative2t1w_xfm, [('T1', 'source_file')]),
+            (fsnative2t1w_xfm, outputnode, [('out_reg_file', 'fsnative2t1w_xfm')]),
+        ])
+        # fmt:on
```
effigies, line 254: This was all that was needed to isolate this step, which drew attention to the fact that we're not mapping skull-strips correctly into fsnative. This should be resolved in another PR.

smriprep/workflows/surfaces.py
```diff
@@ -268,36 +234,131 @@ def init_surface_recon_wf(*, omp_nthreads, hires, name="surface_recon_wf"):
...
+    outputnode = pe.Node(
+        niu.IdentityInterface(
+            fields=[
+                "out_brainmask",
+            ]
+        ),
+        name="outputnode",
+    )
+
+    aseg_to_native_wf = init_segs_to_native_wf()
```
effigies, line 346: Duplicated here and in `init_surface_derivatives_wf`. Acceptably small duplication of effort for conceptual clarity, IMO.

effigies (commented):
@mgxd @oesteban This is passing and ready for review.

I don't think we want to try to squeeze it into 0.10.0, but I've left some comments to help guide a review.

effigies:
@mgxd @oesteban I've reset the merge target to a `next` branch. My plan is to merge this there and make PRs against that. If we need more updates to `master` before the final merge, I'll merge that to `next` as well. Feel free to comment here even after merge (and of course PRs against `next` are welcome).

I will be doing a similar thing over on fMRIPrep.

PR #309, opened 2022-11-09
WRAPPER: Update patch location, use --patch syntax

effigies:
Needed to debug #303

### DIFF
--- a/wrapper/smriprep_docker.py
+++ b/wrapper/smriprep_docker.py
@@ -27,6 +27,7 @@ __credits__ = [
     "Oscar Esteban",
     "Chris Gorgolewski",
     "Christopher J. Markiewicz",
+    "Mathias Goncalves",
     "Russell A. Poldrack",
 ]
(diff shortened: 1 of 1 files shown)


wrapper/smriprep_docker.py
```diff
@@ -222,6 +222,22 @@ def _get_posargs(usage):
...
+                k, v = kv.split("=")
+                d[k] = os.path.abspath(v)
+            setattr(namespace, self.dest, d)
+
+    def _is_file(path, parser):
+        """Ensure a given path exists and it is a file."""
+        path = os.path.abspath(path)
+        if not os.path.isfile(path):
+            raise parser.error("Path should point to a file (or symlink of file): <%s>." % path)
+        return path
```
mgxd, line 240: not currently used?

mgxd (approved):

wrapper/smriprep_docker.py
effigies: Updated to use it. I assume it solved a problem in fmriprep-docker, so might as well have it here...

PR #321, opened 2023-02-03
ENH: make cortical ribbon volume based on signed distance of white/pial surfaces

madisoth:
- Adds interface CreateSignedDistanceVolume for Workbench's wb_command -create-signed-distance-volume

- Adds init_anat_ribbon_wf to surfaces.py, which makes a cortical ribbon volume based on the signed distance function of the white and pial surfaces, as in HCP pipelines. 

- Ribbon volume is output as anat derivative named  "*_desc-ribbon_mask.nii.gz".

### DIFF
--- a/smriprep/__about__.py
+++ b/smriprep/__about__.py
@@ -29,9 +29,7 @@ from ._version import get_versions
 __version__ = get_versions()["version"]
 del get_versions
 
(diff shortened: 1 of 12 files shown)


effigies:
Rebased and merged in #322.

PR #328, opened 2023-03-17
RF: Reflect function name on nipype workflow

mgxd:
Minor (but breaking) change - I spent way too long looking for `morph_grayords_wf` in the working directory...


effigies (approved):
Technically breaking, but I'm inclined to consider it a bug-fix, as it will at worst duplicate a small workflow on re-run, but not invalidate the output of a node that will not get rerun.

I'm sorry about this one...

mgxd:
I'm okay with targeting 0.11.1 if you are 😄

PR #332, opened 2023-04-04
PATCH: Add security fix to vendored jquery

mgxd:
Seemed like the least painful way to address the security concern.

Alternatively, we could upgrade to >= 3.5.0, but I didn't check the difficulty of that transition.


effigies (approved):
I'm fine with this. We can upgrade as well, but it would be nice to see if the javascript starts working first.

oesteban (approved):

PR #337, opened 2023-05-11
ENH: Use ANTs DenoiseImage before conforming anatomical images

### LINKED ISSUE #22: Noisy T1w leads to poor tissue segmentation
image
ds000105/reportlets/fmriprep/sub-5/anat/sub-5_T1w_seg_brainmask.svg

This could be solved by applying a denoising algorithm: https://sites.google.com/site/pierrickcoupe/softwares/denoising-for-medical-imaging/mri-denoising Implemneted as `DenoiseImage` in ANTs

effigies:
Closes #22.

### DIFF
--- a/smriprep/workflows/anatomical.py
+++ b/smriprep/workflows/anatomical.py
@@ -33,7 +33,7 @@ from nipype.interfaces import (
 )
 
 from nipype.interfaces.ants.base import Info as ANTsInfo
-from nipype.interfaces.ants import N4BiasFieldCorrection
+from nipype.interfaces.ants import DenoiseImage, N4BiasFieldCorrection
 
 from niworkflows.engine.workflows import LiterateWorkflow as Workflow
 from niworkflows.interfaces.fixes import FixHeaderApplyTransforms as ApplyTransforms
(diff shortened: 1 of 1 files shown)


mgxd (commented):
I think overall this is a nice addition.

I don't suspect it will add too much (in nibabies it takes about ~4mins) but have you tried to see how it affects the runtime of any test data?

One conceptual question - will this have any effect on confounds?

effigies:
The impact on confounds would be in the definition of white matter/CSF ROIs, if those change. In subjects where the segmentations are kind of salt-and-pepper, we might end up with larger ROIs for those and CompCor.

I did run on some test data and it took ~40sec on 16 cores. So should be <3min on 4 cores.

diramigo:
So the output  of ANTs DenoiseImage is used for recon-all, right? does this mean that using an existing output of recon-all for smriprep could lead to a very different result?

effigies:
This is in 23.1.x. fMRIPrep has always done FreeSurfer a little differently. We have passed INU-corrected T1w images and BOLD masks pre-computed by `antsBrainExtraction.sh` (or, now, a rewrite of that script as a nipype workflow). And for quite some time we've been pre-merging multiple T1w images, rather than leaving that to FreeSurfer.

If you would like to use vanilla FreeSurfer, pre-run it and pass it to fMRIPrep with `--fs-subjects-dir`.

PR #339, opened 2023-05-18
ENH: Project fsLR mesh onto native sphere to enable single-shot resampling into fsLR

### LINKED ISSUE #338: Add white FreeSurfer surface to derivatives; replace smoothwm?
I can't remember the exact reason we chose to use smoothwm instead of white, but white is the fully reconstructed and processed surface. smoothwm is the initial surface post-topological fixing, if I understand it correctly. For all FreeSurfer and Workbench processing (post-recon), white is used, not smoothwm, so we need to provide white to play nicely. Is there any value to continuing to produce smoothwm?

effigies:
This PR does a few things that enable resampling BOLD to fsLR using goodvoxels masking:

1) Uses `wb_command -surface-sphere-project-unproject` to project the fsLR meshes onto the native space (creating what HCP calls `sphere.reg.LR_reg`. If I understand correctly, this is the file that will be replaced if we choose to use MSMSulc in the future.
2) Temporarily adds some atlas files that are not in [tpl-fsLR](https://github.com/templateflow/tpl-fsLR) so we can use them ASAP. The names are kept consistent with HCP naming to avoid confusion. I had to `md5sum` HCP and templateflow files to tell what we already had or didn't have.
3) Adds `wb_command -surface-resample`, although we do not use it here. We use it to generate an fsLR midthickness file, but since it's cheap to generate and we don't save it, it seems more appropriate to use it in fMRIPrep. Putting it here allows us to revise that decision if needed.
4) CHANGES `smoothwm` outputs to `white`. These are needed for workbench operations, and there's no clear reason to keep smoothwm. See #338.

Closes #338.


.circleci/ds005_outputs.txt
```diff
@@ -24,16 +24,20 @@ smriprep/sub-01/anat/sub-01_hemi-L_curv.shape.gii
 smriprep/sub-01/anat/sub-01_hemi-L_inflated.surf.gii
 smriprep/sub-01/anat/sub-01_hemi-L_midthickness.surf.gii
 smriprep/sub-01/anat/sub-01_hemi-L_pial.surf.gii
-smriprep/sub-01/anat/sub-01_hemi-L_smoothwm.surf.gii
+smriprep/sub-01/anat/sub-01_hemi-L_space-fsLR_sphere.surf.gii
+smriprep/sub-01/anat/sub-01_hemi-L_sphere.surf.gii
```
effigies, line 28: Are these sensible names?

smriprep/interfaces/surf.py
```diff
@@ -120,14 +145,42 @@ def normalize_surfs(in_file, transform_file, newpath=None):
         pointset.meta.setdefault("AnatomicalStructureSecondary", "MidThickness")
         pointset.meta.setdefault("GeometricType", "Anatomical")
 
+    # FreeSurfer incorrectly uses "Sphere" for spherical surfaces
+    if pointset.meta.get("GeometricType") == "Sphere":
+        pointset.meta["GeometricType"] = "Spherical"
```
effigies, line 150: This got left in when I decided to make a new interface instead of reusing this one. I could remove it, but it's also not wrong...

effigies (commented):
@mgxd This is ready for review.

.circleci/ds005_outputs.txt
mgxd: Maybe approaching too verbose, but adding `desc-reg` seems useful

```suggestion
smriprep/sub-01/anat/sub-01_hemi-L_space-fsLR_desc-reg_sphere.surf.gii
smriprep/sub-01/anat/sub-01_hemi-L_space-fsaverage_desc-reg_sphere.surf.gii
```

smriprep/interfaces/workbench.py
```diff
@@ -142,3 +142,249 @@ class CreateSignedDistanceVolume(WBCommand):
...
+        argstr="%s",
+        position=2,
+    )
+    sphere_out = traits.File(
+        name_template="%s_unprojected.surf.gii",
+        name_source=["sphere_in"],
+        desc="the output sphere",
+        argstr="%s",
+        position=3,
+    )
```
mgxd, line 237: all mandatory?

smriprep/interfaces/workbench.py
```diff
@@ -142,3 +142,249 @@ class CreateSignedDistanceVolume(WBCommand):
...
+
+
+class SurfaceSphereProjectUnprojectOutputSpec(TraitedSpec):
+    sphere_out = File(desc="the output sphere", exists=True)
```
mgxd, line 241:
```suggestion
    sphere_out = File(desc="the output sphere")
```

smriprep/interfaces/workbench.py
```diff
@@ -142,3 +142,249 @@ class CreateSignedDistanceVolume(WBCommand):
...
+        desc="a relevant surface with <new-sphere> mesh",
+        exists=True,
+        argstr="%s",
+        position=7,
+        requires=['correction_source'],
+    )
+
+
+class SurfaceResampleOutputSpec(TraitedSpec):
+    surface_out = File(desc="the output surface file", exists=True)
```
mgxd, line 368:
```suggestion
    surface_out = File(desc="the output surface file")
```

smriprep/workflows/outputs.py
```diff
@@ -642,6 +644,34 @@ def init_anat_derivatives_wf(
         name="ds_surfs",
         run_without_submitting=True,
     )
+    # Sphere registrations
+    name_regs = pe.MapNode(
```
mgxd, line 648: seems like these are low memory nodes - probably should set mem_gb to something low

mgxd (commented):
This looks good overall, just some minor comments.

One thing - this adds fsLR sphere registration as required step, even on runs that do not wish to do `--cifti-output` / `--project-goodvoxels`. If the processing time this adds is negligible, I'm fine with it, but if not we might want to add someway to avoid this. Or alternative, add some notion of a `--fast-mode` which will avoid non-essential steps.

.circleci/ds005_outputs.txt
effigies: Sounds good. Let's do that.

effigies:
>  If the processing time this adds is negligible, I'm fine with it

5 seconds.

effigies:
@mgxd Okay to merge?

mgxd (approved):

PR #360, opened 2023-08-27
RF: Break up surface workflows for easier mix-and-match in fMRIPrep

effigies:
Should have no effect on current outputs, but:

1) Moves midthickness check/creation back into `surface_recon_wf`, which then needs a sync function to ensure that downstream nodes don't try to extract surfaces before midthickness is present.
2) Split surface and morphometric translation to GIFTI into their own workflows, as these are independent.
3) Pass the selected surfaces and morphometry files as arguments.

Does not need immediate merge. Will probably do more.

### DIFF
--- a/smriprep/workflows/surfaces.py
+++ b/smriprep/workflows/surfaces.py
@@ -223,6 +223,14 @@ gray-matter of Mindboggle [RRID:SCR_002438, @mindboggle].
 
     save_midthickness = pe.Node(nio.DataSink(parameterization=False), name="save_midthickness")
 
+    sync = pe.Node(
+        niu.Function(
+            function=_extract_fs_fields,
+            output_names=['subjects_dir', 'subject_id'],
+        ),
+        name="sync",
+    )
+
     # fmt:off
     workflow.connect([
         # Configuration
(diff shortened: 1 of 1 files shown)


effigies:
This is ready for review.

smriprep/cli/run.py
```diff
@@ -516,9 +527,11 @@ def build_workflow(opts, retval):
             "logging": {"log_directory": str(log_dir), "log_to_file": True},
             "execution": {
                 "crashdump_dir": str(log_dir),
-                "crashfile_format": "txt",
+                "crashfile_format": "pklz",
```
mgxd, line 530: what is the reason for this?

smriprep/interfaces/surf.py
```diff
@@ -42,7 +42,7 @@
 
 class _NormalizeSurfInputSpec(BaseInterfaceInputSpec):
     in_file = File(mandatory=True, exists=True, desc="Freesurfer-generated GIFTI file")
-    transform_file = File(exists=True, desc="FSL or LTA affine transform file")
+    transform_file = File(desc="FSL or LTA affine transform file")
```
mgxd, line 45: Why this change? Seems fine to have a sanity check to ensure the file actually exists

effigies: Ah, right. I was thinking I was making it non-mandatory.

smriprep/cli/run.py
effigies: Likewise, I need the debug and don't want to add flags/config values. We can revert before a release. Can maybe make a list of pre-merge todos?

smriprep/interfaces/surf.py
effigies:
```suggestion
    transform_file = File(exists=True, desc="FSL or LTA affine transform file")
```

effigies:
> will have to port over the MSM sulc stuff once that's in

I think it should be pretty straightforward.

smriprep/cli/run.py
mgxd: Sounds good

PR #369, opened 2023-09-28
MNT: Use f-strings where possible

DimitriPapadopoulos:
Use `.format()` where f-strings would be unreadable.

### DIFF
--- a/smriprep/workflows/anatomical.py
+++ b/smriprep/workflows/anatomical.py
@@ -222,15 +222,11 @@ def init_anat_preproc_wf(
     """
     workflow = Workflow(name=name)
     num_t1w = len(t1w)
-    desc = """
+    desc = f"""
 Anatomical data preprocessing
 
-: """
-    desc += """\
-A total of {num_t1w} T1-weighted (T1w) images were found within the input
-BIDS dataset.""".format(
-        num_t1w=num_t1w
-    )
+: A total of {num_t1w} T1-weighted (T1w) images were found within the input
+BIDS dataset."""
 
     inputnode = pe.Node(
         niu.IdentityInterface(fields=["t1w", "t2w", "roi", "flair", "subjects_dir", "subject_id"]),
--- a/smriprep/__main__.py
+++ b/smriprep/__main__.py
@@ -27,5 +27,5 @@ if __name__ == "__main__":
 
     # `python -m smriprep` typically displays the command as __main__.py
     if "__main__.py" in sys.argv[0]:
-        sys.argv[0] = "%s -m smriprep" % sys.executable
+        sys.argv[0] = f"{sys.executable} -m smriprep"
     main()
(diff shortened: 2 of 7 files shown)


smriprep/workflows/anatomical.py
```diff
@@ -222,15 +222,11 @@ def init_anat_preproc_wf(
...
+    desc = f"""
 Anatomical data preprocessing
 
-: """
-    desc += """\
-A total of {num_t1w} T1-weighted (T1w) images were found within the input
-BIDS dataset.""".format(
-        num_t1w=num_t1w
-    )
+: A total of {num_t1w} T1-weighted (T1w) images were found within the input
```
DimitriPapadopoulos, line 228: The resulting string `desc` is:
```
Anatomical data preprocessing

: A total of num_t1w T1-weighted (T1w) images were found within the input
BIDS dataset.
```
Is this really what we want?

effigies:
I'd rather not merge cosmetic changes in before we merge [next](https://github.com/nipreps/smriprep/tree/next) into master.

Also, `%` is not deprecated functionality, so let's not "upgrade" to `.format()` just because it's newer.

smriprep/workflows/anatomical.py
DimitriPapadopoulos: @effigies What about this one? Was on purpose?

effigies: I think this is a misreading. If you check https://output.circle-artifacts.com/output/job/c1a05c82-b211-4baa-88e9-2b93fcd61dbc/artifacts/0/bids-run/sub-01.html#boilerplate, we do interpolate the number.

DimitriPapadopoulos: I was referring to the leading **': '** at the start of the paragraph. The Mardown outptut looks like this:
```txt
Anatomical data preprocessing

: A total of 1 T1-weighted (T1w) images were found within the input
BIDS dataset.The T1-weighted (T1w) image was corrected for intensity non-uniformity (INU)
```
and as far as I know **': '** does not have a special meaning, and is rendered:
```markdown
Anatomical data preprocessing

: A total of 1 T1-weighted (T1w) images were found within the input
BIDS dataset.The T1-weighted (T1w) image was corrected for intensity non-uniformity (INU)
```

effigies: I think technically it's using [https://garrettgman.github.io/rmarkdown/authoring_pandoc_markdown.html#definition-lists](definition lists) to make "Anatomical data processing" into almost-but-not-quite a section heading.

DimitriPapadopoulos: That was the https://markdownlivepreview.com/ renderer that doesn't understand this Pandoc extension.

Using https://dillinger.io, the Markdown output looks slightly better, but still not perfect (the newline is part of the rendereding):

markdown

Is the Markdown output just for conversion with Pandoc, or is it supposed to be usable as such?

DimitriPapadopoulos:
The idea is that f-strings are faster, not newer.

effigies:
To be clear, I'm fine upgrading to f-strings where it works and is clean. But when we're using `%` instead of `.format()`, it's usually because it's a delayed interpolation and we decided `%` was the cleaner of the two options.

DimitriPapadopoulos:
I understand. In any case, isn't it better to wait for next to be merged into master?

PR #370, opened 2023-09-29
ENH: Add sphere registration to fit workflow, check for precomputed

effigies:
Not much to say. Moved some of the surface conversion into first-order fit, even though they are easy to calculate post hoc because they are needed if msmsulc is used. Initially tried to make some of it conditional, but it made it annoying to construct the complement after the fact. A few extra surfaces in the anatomical derivatives are not our biggest problems, space-wise.

### DIFF
--- a/smriprep/workflows/anatomical.py
+++ b/smriprep/workflows/anatomical.py
@@ -302,9 +302,7 @@ def init_anat_preproc_wf(
     ])
     # fmt:on
     if freesurfer:
-        surfaces = ["white", "pial", "midthickness", "inflated", "sphere_reg", "sphere_reg_fsLR"]
-        if msm_sulc:
-            surfaces.append("sphere_reg_msm")
+        surfaces = ["white", "pial", "midthickness", "inflated"]
 
         surface_derivatives_wf = init_surface_derivatives_wf(
             msm_sulc=msm_sulc,
@@ -553,6 +551,16 @@ BIDS dataset."""
                 "t1w_tpms",
                 "anat2std_xfm",
                 "fsnative2t1w_xfm",
+                # Surface and metric derivatives for fsLR resampling
+                "white",
+                "pial",
+                "midthickness",
+                "sphere",
+                "thickness",
+                "sulc",
+                "sphere_reg",
+                "sphere_reg_fsLR",
+                "sphere_reg_msm",
                 # Reverse transform; not computable from forward transform
                 "std2anat_xfm",
                 # Metadata
@@ -596,7 +604,9 @@ BIDS dataset."""
 
     # Stage 8 results: GIFTI surfaces
     surfaces_buffer = pe.Node(
-        niu.IdentityInterface(fields=["sphere_reg", "sphere", "sulc"]),
+        niu.IdentityInterface(
+            fields=["white", "pial", "midthickness", "sphere", "sphere_reg", "thickness", "sulc"]
+        ),
         name="surfaces_buffer",
     )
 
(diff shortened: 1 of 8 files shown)


effigies:
@mgxd I believe this is ready for review if you've got time.

effigies:
The current failure relates to passing BIDS filenames to Path2BIDS. We expect `rh.thickness_converted.shape.gii` and instead get `sub-01_hemi-R_thickness.shape.gii` because it's being found in the output directory. Need to think about how whether to bypass the extraction or make Path2BIDS more flexible.

effigies:
This is passing. Codecov is complaining about coverage because coverage is bad.

smriprep/workflows/anatomical.py
```diff
@@ -1049,9 +1086,187 @@ def init_anat_fit_wf(
...
+
+    # Stages 8-10: Surface conversion and registration
+    # sphere_reg is needed to generate sphere_reg_fsLR
+    # sphere and sulc are needed to generate sphere_reg_msm
+    # white, pial, midthickness and thickness are needed to resample in the cortical ribbon
+    # TODO: Consider paring down or splitting into a subworkflow that can be called on-demand
+    # A subworkflow would still need to check for precomputed outputs
+    needed_anat_surfs = ["white", "pial", "midthickness"]
+    needed_metrics = ["thickness", "sulc"]
+    needed_spheres = ["sphere_reg", "sphere"]
```
mgxd, line 1101: seems like a good time for sets

```suggestion
    needed_anat_surfs = {"white", "pial", "midthickness"}
    needed_metrics = {"thickness", "sulc"}
    needed_spheres = {"sphere_reg", "sphere"}
```

smriprep/workflows/anatomical.py
```diff
@@ -1049,9 +1086,187 @@ def init_anat_fit_wf(
...
+        for surf in needed_anat_surfs + needed_metrics + needed_spheres
+        if len(precomputed.get(surf, [])) == 2
+    }
+    if found_surfs:
+        LOGGER.info(f"ANAT Stage 8: Found pre-converted surfaces for {list(found_surfs)}")
+        surfaces_buffer.inputs.trait_set(**found_surfs)
+
+    # Stage 8: Surface conversion
+    surfs = [surf for surf in needed_anat_surfs if surf not in found_surfs]
+    spheres = [sphere for sphere in needed_spheres if sphere not in found_surfs]
```
mgxd, line 1115: then we should just be able to do
```suggestion
    surfs = list(needed_anat_surfs - found_surfs.keys())
    spheres = list(needed_spheres - found_surfs.keys())
```

smriprep/workflows/anatomical.py
```diff
@@ -1049,9 +1086,187 @@ def init_anat_fit_wf(
...
+            (gifti_spheres_wf, surfaces_buffer, [
+                (f'outputnode.{sphere}', sphere) for sphere in spheres
+            ]),
+            (sourcefile_buffer, ds_spheres_wf, [("source_files", "inputnode.source_files")]),
+            (gifti_spheres_wf, ds_spheres_wf, [
+                (f'outputnode.{sphere}', f'inputnode.{sphere}') for sphere in spheres
+            ]),
+        ])
+        # fmt:on
+    metrics = [metric for metric in needed_metrics if metric not in found_surfs]
```
mgxd, line 1162:
```suggestion
    metrics = list(needed_metrics - found_surfs.keys())
```

smriprep/workflows/anatomical.py
effigies: Sets can vary in order with each interpreter, so I'd worry about this causing some node to consider its inputs changed. The n is so small here that I think efficiency is not a significant concern.

effigies:
@mgxd It would be nice to depend on this. Is your previous suggestion blocking?

mgxd:
No, it is not blocking

PR #371, opened 2023-10-02
Add a `--fs-reuse-base` option to reuse existing freesurfer outputs from a longitudinal pipeline

bpinsard:
For now, it is not possible to run f/sMRIPrep with existing longitudinal freesurfer output.
If I understood correctly, the current surface generation workflow:
- runs autorecon1 without skullstripping 
- injects a custom (and more robust) `brainmask.mgz` 
- resume to autorecon2/3 without any other tweaks.

Currently, when reusing existing freesurfer outputs, it is not clear to me (maybe docs should be more informative) if the injected brainmask always triggers a rerun of autorecon2/3, from the nipype interface code, I would say no. So it seems only the `brainmask.mgz`  is overwritten, but existing surfaces are directly grabbed.

The simpler way I found to input the base/template recon-all is to use the `FreeSurferSource` and reconnect everything appropriately. Does that make sense, or I missed something (like registering the sMRIPrep generated average to the freesurfer one).


bpinsard:
I am not super familiar with the circleci-codecov setup. Should I add an upload of the `coverage.xml` with a different tag after running with the new option to avoid the errors in the present PR? @effigies 
Thanks for your feedback.

effigies:
Sorry, this got closed due to the merge. I would not worry about codecov at the moment. I'm not sure what you mean about failures, unless you mean the red X's. It looks like it's getting submitted, it's just hard to get good coverage from a nipype workflow.

bpinsard:
No problem. Do you want me to reopen with base `master`? Or this change has been rejected?

PR #375, opened 2023-10-11
FIX: Add surface-modify-sphere call to catch potential sphere elongation

mgxd:
This should address issues raised in https://github.com/nipreps/fmriprep/issues/3092


mgxd:
@effigies here's a branch if you want to test on your data

effigies (approved):
Looks good. Won't really be able to test until it's merged into `next`, but that's fine.

PR #378, opened 2023-10-13
DOC: Fix misspelling found by codespell

DimitriPapadopoulos:


smriprep/interfaces/workbench.py
```diff
@@ -399,7 +399,7 @@ class SurfaceModifySphere(WBCommand):
     a sphere, which can suffer from problems generally not present in
     -surface-sphere-project-unproject.  If the sphere should already be
     centered around the origin, using -recenter may still shift it slightly
-    before changing the radius, which is likely to be undesireable.
+    before changing the radius, which is likely to be undesirable.
```
effigies, line 402: This is actually a copy-paste from connectome workbench.

https://github.com/Washington-University/workbench/blob/c13905b86b054f0ed45071bf180aec394e64202b/src/Algorithms/AlgorithmSurfaceModifySphere.cxx#L57

I'm fine fixing it here, but we'll generally pick up their misspellings without critique when we create these interfaces.

effigies (approved):

PR #383, opened 2023-11-08
FIX: Invert sulcal depth metric before passing to MSM, use HCP atlas files

effigies:
This PR narrowly inverts sulcal depth metrics before passing to MSM. I do not currently invert sulcal depth and curvature for GIFTI/CIFTI morphometric outputs.


mgxd (approved):
I think this is right, but let's wait on CI to catch any smoke. One thing we should probably discuss (probably at the next nipreps meeting) is the updating of templateflow templates, and the migration of these templates into it.

And thanks again to @madisoth for tracking this down!

PR #386, opened 2023-11-15
ENH: Annotate mris_expand with thread usage

effigies:
This one's a little weird, due to the multiple inheritance and also the inefficiency of multithreading in `mris_expand`. 

I noticed `midthickness` taking a long time while running tests, and I was pretty sure I'd seen that it could be parallelized. 

From watching `htop` and running `mris_expand` with a set of `OMP_NUM_THREADS` values, I see that generally you get about 65-75% core utilization, generally a bit under. We therefore overallocate threads by about 50% to make the core usage match the desired number of threads, occasionally peaking over but mostly sitting a bit under. This works pretty well for 2-12 desired threads (3-18 `OMP_NUM_THREADS`), but seems almost completely limited after that, with 24 threads utilizing maybe 13 cores at best. Therefore, I set this to max out at 12 for scheduling purposes.

The memory usage of this tool is low and does not seem to increase noticeably with number of threads.

If `n_procs == 1`, `str(1 * 3 // 2) == '1'`, so this does not interfere with people attempting to run things single-threaded.

### DIFF
--- a/smriprep/interfaces/freesurfer.py
+++ b/smriprep/interfaces/freesurfer.py
@@ -293,12 +293,20 @@ class MRIsConvertData(fs.utils.MRIsConvert):
         return super()._gen_filename(name)
 
 
-class MakeMidthickness(nwfs.MakeMidthickness):
+class MakeMidthicknessInputSpec(nwfs._MakeMidthicknessInputSpec, fs.base.FSTraitedSpecOpenMP):
+    pass
+
+
+class MakeMidthickness(nwfs.MakeMidthickness, fs.base.FSCommandOpenMP):
     """Patched MakeMidthickness interface
 
     Ensures output filenames are specified with hemisphere labels, when appropriate.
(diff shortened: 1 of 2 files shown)


smriprep/interfaces/freesurfer.py
```diff
@@ -316,9 +324,18 @@ class MakeMidthickness(nwfs.MakeMidthickness):
...
+
     def _format_arg(self, name, trait_spec, value):
         # FreeSurfer at some point changed whether it would add the hemi label onto the
         # surface. Therefore we'll do it ourselves.
         if name == "out_name":
             value = self._associated_file(self.inputs.in_file, value)
         return super()._format_arg(name, trait_spec, value)
+
+    def _num_threads_update(self):
+        """mris_expand"""
```
effigies, line 337:
```suggestion
```

effigies:
If there are no objections, I'll merge this before a release. It should reduce a 20min job to 2.5-4 minutes.

PR #280, opened 2022-04-26
ENH: Add FastSurfer file detection in subjects_dir, draft interface (WIP)

pcamach2:
_Supports #278_ 

Adds FastSurfer detection utility incorporated to smriprep.workflows.anatomical.py to fix failure when providing FastSurfer subjects dir as existing Freesurfer subjects dir. This utility `check_fastsurfer` employs a fix of touching a labels intensity file expected by Freesurfer, but not currently produced by FastSurfer. FastSurfer outputs from running with the `--surfreg` flag can then be used to successfully complete the sMRIPrep workflow.

Adds draft of basic interface for FastSurfer to smriprep interfaces, which work on my test datasets when running on the modified Docker or Singularity images (see f8972dac90f75d7e051723b19e6bb826f1cb696e for changes to Dockerfile).

To support the using the FastSurfer interface/wrapper to run FastSurfer as part of sMRIPrep, this PR also adds the following argument to the cli (https://github.com/pcamach2/smriprep/blob/pcamach2-fastsurfer-patch-dev/smriprep/cli/run.py)

```
    g_surfs_xor.add_argument(
        "--fastsurfer-recon",
        action="store_true",
        dest="run_fastsurfer",
        help="enable FastSurfer surface preprocessing.",
    )
```

As mentioned in the above issue, I am wondering where best to split workflow behavior (in anatomical.py?) based on the above argument. Would it be better to make separate surfaces.py versions for FastSurfer and FreeSurfer or just different workflows within a unified surfaces.py? Note that if a dataset includes FLAIR or T2w, FreeSurfer will still want to process them automatically even if there is FastSurfer output (this adds on a few hours).

PR #393, opened 2023-11-21
Add --fs-no-resume option to reuse existing freesurfer outputs without resuming (eg. longitudinal pipeline base)

bpinsard:
(Reopen of #371)

For now, it is not possible to run f/sMRIPrep with existing longitudinal freesurfer output.
If I understood correctly, the current surface generation workflow:

runs autorecon1 without skullstripping
injects a custom (and more robust) brainmask.mgz
resume to autorecon2/3 without any other tweaks.

The simpler way I found to input the base/template recon-all is to use the FreeSurferSource and reconnect everything appropriately. Does that make sense, or I missed something (like registering the sMRIPrep generated average to the freesurfer one).


effigies (commented):
Okay, getting back to this. Thanks for the patience. As I understand it, instead of trying to correctly wrap longitudinal FreeSurfer in the same way we've been wrapping single-session FreeSurfer, this just give the user a "trust me, I know what I'm doing" button.

If that's the case, I think the better approach is probably to make the flag `--fs-no-resume` or something like that (open to alternatives), which would allow it to work with any FreeSurfer-like directory structure. This would work equally well with longitudinal FreeSurfer, fastsurfer (#280) or any other technique that comes out.

One thing that would be worth considering in that case is to provide a function to check for necessary files and error out early if they are not found. For example we need `{lh,rh}.{white,pial,thickness,sulc,curv,sphere,sphere.reg}` as well as `T1.mgz`, `aparc.mgz`, `aparc+aseg.mgz` and either `FLAIR.mgz` or `T2.mgz`, if T2w/FLAIR are detected.

That said, since this would be a "power user" feature, I don't think it's critical to write a hand-holder on the first pass.

bpinsard:
> Okay, getting back to this. Thanks for the patience. As I understand it, instead of trying to correctly wrap longitudinal FreeSurfer in the same way we've been wrapping single-session FreeSurfer, this just give the user a "trust me, I know what I'm doing" button.
Exactly, wrapping longitudinal pipeline is more complex. 

> If that's the case, I think the better approach is probably to make the flag `--fs-no-resume` or something like that (open to alternatives), which would allow it to work with any FreeSurfer-like directory structure. This would work equally well with longitudinal FreeSurfer, fastsurfer (#280) or any other technique that comes out.
Happy with renaming the flag, I had not thought about the fastsurfer usecase but that's a very good point.

> One thing that would be worth considering in that case is to provide a function to check for necessary files and error out early if they are not found. For example we need `{lh,rh}.{white,pial,thickness,sulc,curv,sphere,sphere.reg}` as well as `T1.mgz`, `aparc.mgz`, `aparc+aseg.mgz` and either `FLAIR.mgz` or `T2.mgz`, if T2w/FLAIR are detected.
Such as a new nipype node that checks that the traits are not undefined?

> That said, since this would be a "power user" feature, I don't think it's critical to write a hand-holder on the first pass.

smriprep/workflows/surfaces.py
```diff
@@ -239,39 +247,82 @@ def init_surface_recon_wf(
         name='sync',
     )
 
+    if not fs_no_resume:
```
effigies, line 250: Large parts of this block are duplicated from above.

1) Create all the nodes and only stick connect statements in the `if`/`else` block
2) Create all the unconditional nodes outside the `if`/`else` block, and the conditional nodes inside

bpinsard:
Interestingly, https://github.com/templateflow/tpl-fsaverage/pull/5 broke the CI run here (and likely any smriprep run?)
https://github.com/nipreps/smriprep/blob/master/smriprep/cli/run.py#L416 should be fixed? or niworkflows.utils.misc._copy_any ?
@oesteban

effigies:
I guess we want

https://github.com/nipreps/smriprep/blob/45c2e2fc088839979f18b83be630a6f9d515b416/smriprep/cli/run.py#L416

to become:

```python
            dseg_tsv = str(api.get('fsaverage', hemi=None, seg='aparc', suffix='dseg', extension=['.tsv']))
```

bpinsard:
> I guess we want
> 
> https://github.com/nipreps/smriprep/blob/45c2e2fc088839979f18b83be630a6f9d515b416/smriprep/cli/run.py#L416
> 
> to become:
> 
> ```python
>             dseg_tsv = str(api.get('fsaverage', hemi=None, seg='aparc', suffix='dseg', extension=['.tsv']))
> ```

Opened #420 
I guess this should be back-ported to the LTS as it will also pull from the latest templateflow version.

smriprep/workflows/surfaces.py
```diff
@@ -239,39 +247,64 @@ def init_surface_recon_wf(
...
+                ('outputnode.subject_id', 'subject_id'),
+            ]),
+            (autorecon_resume_wf, save_midthickness, [
+                ('outputnode.subjects_dir', 'base_directory'),
+                ('outputnode.subject_id', 'container'),
+            ]),
+        ])
+        # fmt:on
+    else:
+        fs_base_inputs = autorecon1 = pe.Node(nio.FreeSurferSource(), name='fs_base_inputs')
```
effigies, line 288:
```suggestion
        # Pretend to be the autorecon1 node so fsnative2t1w_xfm gets run ASAP
        fs_base_inputs = autorecon1 = pe.Node(nio.FreeSurferSource(), name='fs_base_inputs')
```

smriprep/workflows/surfaces.py
```diff
@@ -239,39 +247,64 @@ def init_surface_recon_wf(
...
+            (autorecon_resume_wf, get_surfaces, [
+                ('outputnode.subjects_dir', 'subjects_dir'),
+                ('outputnode.subject_id', 'subject_id'),
+            ]),
+            (autorecon_resume_wf, save_midthickness, [
+                ('outputnode.subjects_dir', 'base_directory'),
+                ('outputnode.subject_id', 'container'),
+            ]),
+        ])
+        # fmt:on
```
effigies, line 286:
```suggestion
        workflow.connect([
            # Configuration
            (inputnode, recon_config, [('t1w', 't1w_list'),
                                       ('t2w', 't2w_list'),
                                       ('flair', 'flair_list')]),
            # Passing subjects_dir / subject_id enforces serial order
            (inputnode, autorecon1, [('subjects_dir', 'subjects_dir'),
                                     ('subject_id', 'subject_id')]),
            (autorecon1, skull_strip_extern, [('subjects_dir', 'subjects_dir'),
                                              ('subject_id', 'subject_id')]),
            (skull_strip_extern, autorecon_resume_wf, [('subjects_dir', 'inputnode.subjects_dir'),
                                                       ('subject_id', 'inputnode.subject_id')]),
            # Reconstruction phases
            (inputnode, autorecon1, [('t1w', 'T1_files')]),
            (inputnode, fov_check, [('t1w', 'in_files')]),
            (fov_check, autorecon1, [('out', 'flags')]),
            (recon_config, autorecon1, [('t2w', 'T2_file'),
                                        ('flair', 'FLAIR_file'),
                                        ('hires', 'hires'),
                                        # First run only (recon-all saves expert options)
                                        ('mris_inflate', 'mris_inflate')]),
            (inputnode, skull_strip_extern, [('skullstripped_t1', 'in_brain')]),
            (recon_config, autorecon_resume_wf, [('use_t2w', 'inputnode.use_T2'),
                                                 ('use_flair', 'inputnode.use_FLAIR')]),
            # Generate mid-thickness surfaces
            (autorecon_resume_wf, get_surfaces, [
                ('outputnode.subjects_dir', 'subjects_dir'),
                ('outputnode.subject_id', 'subject_id'),
            ]),
            (autorecon_resume_wf, save_midthickness, [
                ('outputnode.subjects_dir', 'base_directory'),
                ('outputnode.subject_id', 'container'),
            ]),
        ])  # fmt:skip
```

smriprep/workflows/surfaces.py
```diff
@@ -239,39 +247,64 @@ def init_surface_recon_wf(
...
+                ('subject_id', 'subject_id'),
+            ]),
+            (inputnode, save_midthickness, [
+                ('subjects_dir', 'base_directory'),
+                ('subject_id', 'container'),
+            ]),
+        ])
+        # fmt:on
+
+    # fmt:off
```
effigies, line 306:
```suggestion
        workflow.connect([
            (inputnode, fs_base_inputs, [
                ('subjects_dir', 'subjects_dir'),
                ('subject_id', 'subject_id'),
            ]),
            # Generate mid-thickness surfaces
            (inputnode, get_surfaces, [
                ('subjects_dir', 'subjects_dir'),
                ('subject_id', 'subject_id'),
            ]),
            (inputnode, save_midthickness, [
                ('subjects_dir', 'base_directory'),
                ('subject_id', 'container'),
            ]),
        ])  # fmt:skip

```

effigies (approved):
LGTM. Added a comment and updated the `fmt:<on|off|skip>` directives.

PR #394, opened 2023-11-21
MNT: update target Python version for black

DimitriPapadopoulos:
Match required Python version, which is currently 3.10:
https://github.com/nipreps/smriprep/blob/a3df930f9fa84f1a8a1bd1ebb35ad2a1c0fe6348/pyproject.toml#L20


effigies:
Should actually just remove it, since black uses `requires-python` if present.

This is part of the branch I'm working on, so I'm going to close this.

PR #418, opened 2024-03-06
FIX: Fetch templates during workflow construction

mgxd:
This addresses https://github.com/nipreps/fmriprep/issues/3240, a few changes were necessary:

- Fetch templates during workflow creation - without this, I was still observing race conditions causing the T1w/brain mask files to be empty (not yet gotten through `templateflow.api.get()`)
- Add `sloppy` parameter to `init_template_iterator_wf`, since the T1w/brain mask's resolution (used in [`SpatialNormalization`](https://github.com/nipreps/niworkflows/blob/6edfb38f1aa6c706c8a4942ebadd9041738978b9/niworkflows/interfaces/norm.py#L130)) depends on this parameter as well.

I stripped out most of the logic in `TemplateFlowSelect` to avoid duplicating the behavior, since it is essentially linked.

### DIFF
--- a/smriprep/workflows/anatomical.py
+++ b/smriprep/workflows/anatomical.py
@@ -974,9 +974,9 @@ the brain-extracted T1w using `fast` [FSL {fsl_ver}, RRID:SCR_002823, @fsl_fast]
     templates = []
     found_xfms = {}
     for template in spaces.get_spaces(nonstandard=False, dim=(3,)):
-        from ..utils.misc import get_template_t1w
+        from smriprep.interfaces.templateflow import fetch_template_files
 
-        get_template_t1w(template, sloppy)
+        fetch_template_files(template, specs=None, sloppy=sloppy)
         xfms = precomputed.get('transforms', {}).get(template, {})
         if set(xfms) != {'forward', 'reverse'}:
             templates.append(template)
--- a/smriprep/workflows/anatomical.py
+++ b/smriprep/workflows/anatomical.py
@@ -974,9 +974,6 @@ the brain-extracted T1w using `fast` [FSL {fsl_ver}, RRID:SCR_002823, @fsl_fast]
     templates = []
     found_xfms = {}
     for template in spaces.get_spaces(nonstandard=False, dim=(3,)):
-        from smriprep.interfaces.templateflow import fetch_template_files
-
-        fetch_template_files(template, specs=None, sloppy=sloppy)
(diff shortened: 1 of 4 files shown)


smriprep/workflows/anatomical.py
```diff
@@ -972,6 +973,9 @@ def init_anat_fit_wf(
     templates = []
     found_xfms = {}
     for template in spaces.get_spaces(nonstandard=False, dim=(3,)):
+        from smriprep.interfaces.templateflow import fetch_template_files
+
+        fetch_template_files(template, specs=None, sloppy=sloppy)
```
effigies, line 978: This feels like the wrong place to put this. Could we instead put it in the workflows that use `TemplateFlowSelect`, to more tightly tie the places it's needed?

mgxd: where would you suggest we put it? The reason I chose this spot is it is hit regardless of whether derivatives are found, and didn't want to duplicate this call across workflows.

effigies: Well I was thinking that it's okay to duplicate, to make sure that it's done however workflows are mixed and matched.

mgxd: [e890872](https://github.com/nipreps/smriprep/pull/418/commits/e890872e4594ac590546dba23b1e9bc362969a3a) moves this into `init_template_iterator_wf`

smriprep/workflows/outputs.py
```diff
@@ -1121,6 +1126,11 @@ def init_template_iterator_wf(*, spaces, name='template_iterator_wf'):
 
     The fields in `outputnode` can be used as if they come from a single template.
     """
+    for template in spaces.get_spaces(nonstandard=False, dim=(3,)):
+        from smriprep.interfaces.templateflow import fetch_template_files
+
```
effigies, line 1131:
```suggestion
    for template in spaces.get_spaces(nonstandard=False, dim=(3,)):
```

smriprep/workflows/outputs.py
```diff
@@ -34,10 +34,13 @@
 from ..interfaces import DerivativesDataSink
 from ..interfaces.templateflow import TemplateFlowSelect
```
effigies, line 35:
```suggestion
from ..interfaces.templateflow import TemplateFlowSelect, fetch_template_files
```

effigies (approved):

PR #420, opened 2024-03-20
[FIX] Require recent templateflow, select correct aparc dseg.tsv

bpinsard:
see https://github.com/nipreps/smriprep/pull/393#issuecomment-2010268610


effigies:
What a mess. I think I know the issue. Let me push to your branch...

effigies (approved):

PR #433, opened 2024-04-29
RF: Remove much of the hardcoded `T1w` fields in favor of `anat`

mgxd:
This is a supplemental PR to https://github.com/nipreps/nibabies/pull/360 - since nibabies also relies on smriprep, but may be processing in T2w space, this PR generalize many of the workflows/nodes/fields to account for other anatomical references.

### DIFF
--- a/smriprep/workflows/anatomical.py
+++ b/smriprep/workflows/anatomical.py
@@ -287,7 +287,6 @@ def init_anat_preproc_wf(
     ds_std_volumes_wf = init_ds_anat_volumes_wf(
         bids_root=bids_root,
         output_dir=output_dir,
-        name='ds_std_volumes_wf',
     )
 
     workflow.connect([
@@ -320,10 +319,10 @@ def init_anat_preproc_wf(
         ]),
         (anat_fit_wf, ds_std_volumes_wf, [
             ('outputnode.t1w_valid_list', 'inputnode.source_files'),
-            ('outputnode.t1w_preproc', 'inputnode.t1w_preproc'),
-            ('outputnode.t1w_mask', 'inputnode.t1w_mask'),
-            ('outputnode.t1w_dseg', 'inputnode.t1w_dseg'),
-            ('outputnode.t1w_tpms', 'inputnode.t1w_tpms'),
+            ('outputnode.t1w_preproc', 'inputnode.anat_preproc'),
+            ('outputnode.t1w_mask', 'inputnode.anat_mask'),
+            ('outputnode.t1w_dseg', 'inputnode.anat_dseg'),
+            ('outputnode.t1w_tpms', 'inputnode.anat_tpms'),
         ]),
         (template_iterator_wf, ds_std_volumes_wf, [
             ('outputnode.std_t1w', 'inputnode.ref_file'),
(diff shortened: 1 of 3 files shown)


mgxd:
@effigies This should be ready for review. I've tried to limit any breaking changes, but regardless this will need to be included in a new minor release

smriprep/workflows/anatomical.py
```diff
@@ -1375,7 +1365,7 @@ def init_anat_template_wf(
     longitudinal: bool,
     omp_nthreads: int,
     num_files: int,
-    contrast: str,
+    image_type: ty.Literal['T1w', 'T2w'] = 'T1w',
```
effigies, line 1368: Does this need a default for some reason?

smriprep/workflows/outputs.py
```diff
@@ -538,44 +571,44 @@ def init_ds_template_registration_wf(
...
 
-    ds_t1w2std_xfm = pe.MapNode(
+    ds_anat2std_xfm = pe.MapNode(
         DerivativesDataSink(
             base_directory=output_dir,
             mode='image',
             suffix='xfm',
             dismiss_entities=['desc'],
-            **{'from': 'T1w'},
+            **{'from': image_type},
```
effigies, line 593: We should think about whether the subject-native space should continue to be called `T1w`. We could call it `anat`, so we'd have `from-anat_to-MNI_xfm.h5` and `space-anat_desc-preproc_bold.nii.gz`.

But that does not need to be decided now, and it's better not to change the output names until we're sure that's what we want.

smriprep/workflows/outputs.py
```diff
@@ -731,6 +765,8 @@ def init_ds_surfaces_wf(
                 ds_surf.inputs.desc = 'reg'
                 if surf == 'sphere_reg_fsLR':
                     ds_surf.inputs.space = 'fsLR'
+                elif surf == 'sphere_reg_dhcpAsym':
+                    ds_surf.inputs.space = 'dhcpAsym'
```
effigies, line 769: Are you using msmsulc to align to dhcpAsym, or is that what MCRIBs registers to by default?

effigies (approved):
LGTM. Left some non-blocking comments.

smriprep/workflows/anatomical.py
mgxd: I was intending to minimize changes in fmriprep

smriprep/workflows/outputs.py
mgxd: MCRIBs already produces the registration to dhcpAsym, which makes it easier to get to CIFTI space.

We aren't running msmsulc yet, mostly due to my lack of clarity of the differences between dhcpAsym / fsLR surfaces.

smriprep/workflows/anatomical.py
effigies: I'd rather rip off the bandaid. Anything that you think should be generic, let's make it generic and deal with it now.

smriprep/workflows/outputs.py
mgxd: I like calling it `anat`, but I feel it still carries some ambiguity. Perhaps having some metadata (either in the sidecar or a separate, generated file) covering exactly this space is (image dimensions, files used to derive it, etc) may be helpful.

smriprep/workflows/anatomical.py
mgxd: all set - probably can squash all of this when merging

PR #440, opened 2024-06-03
FIX: Avoid datasink connection, which will always invalidate cache

mgxd:
Any `ds_*` workflow should probably not have an `outputnode` to avoid this problem.


effigies:
The goal really is to use the outputs of the derivatives workflows, so that we're using the same inputs as downstream tools (or reruns with these derivatives as outputs) would be.

I think the problem here is with the complexity of `DerivativesDataSink`, which can actually modify the contents of files (e.g., setting the right dtypes). I believe the sink nodes already do not make copies when the input matches the output, but the fixup breaks that check. So maybe we should split into prepare and sink nodes, where the prepare node just passes through the filename if no changes are to be made?

mgxd:
I agree - this boils down to splitting up the DDS behemoth into much more succinct parts. Should I close this?

effigies:
No. :-)

PR #448, opened 2024-07-31
RF: Load package data with acres

mgxd:


smriprep/data/__init__.py
```diff
@@ -1,25 +1,3 @@
-import atexit
-from contextlib import ExitStack
-from pathlib import Path
+from acres import Loader
```
effigies, line 1:
```suggestion
"""sMRIPRep data files

.. autofunction:: load

.. automethod:: load.readable

.. automethod:: load.as_path

.. automethod:: load.cached
"""

from acres import Loader
```

smriprep/workflows/surfaces.py
```diff
@@ -56,7 +56,7 @@
 from smriprep.interfaces.surf import MakeRibbon
 from smriprep.interfaces.workbench import SurfaceResample
 
-from ..data import load_resource
+from ..data import load
```
effigies, line 59: Maybe

```suggestion
from .. import data
```

And use `data.load()` below? I feel like a bare `load` variable is really ambiguous. I almost want to say `import smriprep` and use `smriprep.data.load()`.

mgxd: Agreed, I think it's much cleaner.

I went with importing and aliasing `smriprep.data.load` to be used like:
`smriprep.load_data(...)`

effigies (approved):

PR #450, opened 2024-08-20
FIX: Select function in segmentation resampling workflow

mgxd:
The segmentation selection code was a bit convoluted, this addresses the following:

- Adding a mapping for common replacements
- Avoiding string iteration
- Stopping iteration over files upon the first match
- Raising FileNotFoundError, rather than an unclear IndexError, if no matches are found.

### DIFF
--- a/smriprep/workflows/surfaces.py
+++ b/smriprep/workflows/surfaces.py
@@ -1226,20 +1226,8 @@ def init_segs_to_native_wf(
         name='resample',
     )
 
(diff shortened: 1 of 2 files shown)


mgxd:
@effigies can we sneak this into a patch release, or will it require a new minor?

effigies:
I'm okay with a patch release. The workflow cache invalidation is straightforward here.

PR #451, opened 2024-09-04
ENH: Calculate T1/T2 ratio

effigies:
Was hoping to get this all ready in a single shot, but I won't be able to work on this more for a couple weeks at least.


effigies:
The status of this is that this seems to be an unreliable method without a prescribed method for normalizing T1w and T2w intensities.

https://doi.org/10.3389/fnhum.2014.00671 has one method for normalizing ("calibrating") using non-brain tissue masks to identify reference T1w and T2w intensities and rescaling to match these values for a template. This is probably beyond the scope of smriprep, but could be done as a post-workflow.

PR #457, opened 2024-12-11
ENH: Add `t2w_file` output to `TemplateFlowSelect`

mgxd:
If found, great. If not, just returns `Undefined`.

### DIFF
--- a/smriprep/interfaces/templateflow.py
+++ b/smriprep/interfaces/templateflow.py
@@ -30,6 +30,7 @@ from nipype.interfaces.base import (
     InputMultiObject,
     SimpleInterface,
     TraitedSpec,
+    Undefined,
     isdefined,
     traits,
 )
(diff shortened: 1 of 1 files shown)


smriprep/interfaces/templateflow.py
```diff
@@ -206,4 +212,6 @@ def fetch_template_files(
     files['mask'] = tf.get(name[0], desc='brain', suffix='mask', **specs) or tf.get(
         name[0], label='brain', suffix='mask', **specs
     )
+    # Not guaranteed to exist so add fallback
+    files['t2w'] = tf.get(name[0], desc=None, suffix='T2w', **specs) or Undefined
```
effigies, line 216: Is it guaranteed to be only one file?

effigies (approved):
Fine with this, since it looks the same as the lines just above. Posted question that I hope isn't a rabbit hole pre-review.

smriprep/interfaces/templateflow.py
mgxd: nope, and neither are the above 😄  - should i add a check?

mgxd: for the sake of progress, we'll keep an eye out and address accordingly

PR #458, opened 2024-12-11
chore: Adopt src/ layout and tox

effigies:


effigies:
@mgxd if you could have a look, I'd appreciate it.

mgxd (approved):
Looks really good - one question since it's not clear to me whether its present:

Do we still compare the tag version with the built version of the package? I found that check useful for catching mislabeled and dirty releases

effigies:
The tests for the docker image are still there on Circle, which is where we've had actual problems AFAIR. Building the wheels for PyPI is generally less problematic, since you clone the whole repo and there isn't anything like `.dockerignore` to get in the way.

PR #462, opened 2024-12-13
fix: Restore init_resample_midthickness_wf for backwards compatibility

effigies:
Sorry @mgxd. Try out this branch?

### DIFF
--- a/src/smriprep/workflows/surfaces.py
+++ b/src/smriprep/workflows/surfaces.py
@@ -1316,13 +1316,55 @@ def init_anat_ribbon_wf(name='anat_ribbon_wf'):
     return workflow
 
 
+def init_resample_midthickness_wf(
+    grayord_density: ty.Literal['91k', '170k'],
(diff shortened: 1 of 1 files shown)


mgxd:
@effigies I'm fine with using https://github.com/nipreps/smriprep/tree/dev-nibabies for the time being - will shift back to `master` shortly after a nibabies release

effigies:
@mgxd The plan is just to pin to a dev branch, or do you need a release without that change in? If you're pushing to pypi, you'll presumably need a PyPI release.

mgxd:
i just pushed an [alpha to docker hub](https://hub.docker.com/layers/nipreps/nibabies/25.0.0a1/images/sha256-322bdacc63101b3a97b470f0f4bc20f710995fc610c09d3cd34c1256a66e1cd2) to get hbcd processing underway - plan is to sync up before the 25.0.0 release

effigies:
Got it.

PR #472, opened 2025-04-30
FIX: set fast bias_iters parameter to 0

### LINKED ISSUE #471: current fast node still runs bias field correction even if it does not remove the field
**Describe the bug**
the fast node currently implements the following command line:
`fast -N -p -g -S 1 [anatomical image]`. Note the flag: `-N` which runs fast  with the option to not remove the bias field. However, the flag for the number of main-loop iterations (`-I`) during bias-field removal is absent so remains unchanged from the default (4).

The nipype interface [currently](https://nipype.readthedocs.io/en/1.10.0/api/generated/nipype.interfaces.fsl.preprocess.html#fast) requires an integer => 1, so it's not as simple as currently setting `bias_iters = 0`.

An updated interface was applied in fmriprep-rodents at nipreps/fmriprep-rodents@d4caa4f9fa971056be2bba7145b8576f33c22c0c which I can upstream to sMRIPrep

**Additional context**
This might have implications for #109

eilidhmacnicol:
This adds a new fsl interface module, which contains a patch for nipype's fast interface. The nipype interface requires bias_iters to be `1 <= integer <= 10`, but setting `bias_iters=0` is required to bypass bias field correction.

The patched interface is dropped into the anatomical workflow and the parameters are accordingly updated.

closes #471

### DIFF
--- a/src/smriprep/interfaces/fsl.py
+++ b/src/smriprep/interfaces/fsl.py
@@ -1,5 +1,6 @@
 from nipype.interfaces.base import traits
-from nipype.interfaces.fsl.preprocess import FAST as _FAST, FASTInputSpec
+from nipype.interfaces.fsl.preprocess import FAST as _FAST
+from nipype.interfaces.fsl.preprocess import FASTInputSpec
 
 
 class _FixTraitFASTInputSpec(FASTInputSpec):
@@ -24,4 +25,3 @@ class FAST(_FAST):
     """
 
     input_spec = _FixTraitFASTInputSpec
-
--- a/src/smriprep/interfaces/fsl.py
+++ b/src/smriprep/interfaces/fsl.py
@@ -21,8 +21,6 @@ class FixBiasItersFAST(FAST):
 
     def _run_interface(self, runtime, correct_return_codes=(0,)):
         # Run normally
-        runtime = super()._run_interface(
-            runtime, correct_return_codes
-        )
+        runtime = super()._run_interface(runtime, correct_return_codes)
 
         return runtime
--- a/src/smriprep/interfaces/fsl.py
+++ b/src/smriprep/interfaces/fsl.py
@@ -17,11 +17,12 @@ class FAST(_FAST):
     A replacement for nipype.interfaces.fsl.preprocess.FAST that allows
     `bias_iters=0` to disable bias field correction entirely
 
-    >>> from smriprep.interfaces.fsl import FixBiasItersFAST as FAST
-    >>> fast = fsl.FAST()
+    >>> from smriprep.interfaces.fsl import FAST
+    >>> fast = FAST()
     >>> fast.inputs.in_files = 'sub-01_desc-warped_T1w.nii.gz'
+    >>> fast.inputs.bias_iters = 0
     >>> fast.cmdline
-    'fast -o fast_ -S 1 -I 0 sub-01_desc-warped_T1w.nii.gz'
+    'fast -I 0 -S 1 sub-01_desc-warped_T1w.nii.gz'
     """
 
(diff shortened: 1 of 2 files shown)


eilidhmacnicol:
@effigies @mgxd @oesteban do I need any tests for this interface?

src/smriprep/interfaces/fsl.py
```diff
@@ -0,0 +1,26 @@
...
+    `bias_iters=0` to disable bias field correction entirely
+    """
+
+    input_spec = _FixTraitFASTInputSpec
+
+    def _run_interface(self, runtime, correct_return_codes=(0,)):
+        # Run normally
+        runtime = super()._run_interface(runtime, correct_return_codes)
+
+        return runtime
```
mgxd, line 26: Inheritance will take care of this
```suggestion
```

src/smriprep/interfaces/fsl.py
```diff
@@ -0,0 +1,26 @@
...
+class _FixTraitFASTInputSpec(FASTInputSpec):
+    bias_iters = traits.Range(
+        low=0,
+        high=10,
+        argstr='-I %d',
+        desc='number of main-loop iterations during bias-field removal',
+    )
+
+
+class FixBiasItersFAST(FAST):
```
mgxd, line 14: Since this is supported by FAST, no need to call it anything else. We should probably backport this to nipype at some point...
```suggestion
class FAST(_FAST):
```

src/smriprep/interfaces/fsl.py
```diff
@@ -0,0 +1,26 @@
+from nipype.interfaces.base import traits
+from nipype.interfaces.fsl.preprocess import FAST, FASTInputSpec
```
mgxd, line 2:
```suggestion
from nipype.interfaces.fsl.preprocess import _FAST, FASTInputSpec
```

src/smriprep/workflows/anatomical.py
```diff
@@ -57,6 +57,7 @@
 import smriprep
 
 from ..interfaces import DerivativesDataSink
+from ..interfaces.fsl import FixBiasItersFAST as FAST
```
mgxd, line 60:
```suggestion
from ..interfaces.fsl import FAST
```

mgxd (commented):
I think adding a doctest explicitly setting `bias_iters` to 0 will suffice

https://github.com/nipy/nipype/blob/8234ec318489930fa17487cf15d210420214b00a/nipype/interfaces/fsl/preprocess.py#L378-L383

src/smriprep/interfaces/fsl.py
effigies: @mgxd you mean?

```suggestion
from nipype.interfaces.fsl.preprocess import FAST as _FAST, FASTInputSpec
```

src/smriprep/interfaces/fsl.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+    A replacement for nipype.interfaces.fsl.preprocess.FAST that allows
+    `bias_iters=0` to disable bias field correction entirely
+
+    >>> from smriprep.interfaces.fsl import FAST
+    >>> fast = FAST()
+    >>> fast.inputs.in_files = 'sub-01_desc-warped_T1w.nii.gz'
+    >>> fast.inputs.bias_iters = 0
+    >>> fast.cmdline
+    'fast -S 1 -I 0 sub-01_desc-warped_T1w.nii.gz'
```
effigies, line 25:
```suggestion
    'fast -I 0 -S 1 sub-01_desc-warped_T1w.nii.gz'
```

PR #476, opened 2025-05-02
ENH: Tag structural workflows

mgxd:


src/smriprep/workflows/surfaces.py
```diff
@@ -835,6 +839,7 @@ def init_msm_sulc_wf(*, sloppy: bool = False, name: str = 'msm_sulc_wf'):
     return workflow
 
 
+@tag('anat.gifti-surfaces')
 def init_gifti_surfaces_wf(
     *,
     surfaces: list[str] = ('pial', 'midthickness', 'inflated', 'white'),
```
effigies, line 845: This workflow can be instantiated multiple times in a subject's workflow, e.g., with anatomical surfaces and with spheres. Tagging this here as opposed to after calling would mean that attempting to splice would replace two workflows with one.

src/smriprep/workflows/surfaces.py
```diff
@@ -932,6 +937,7 @@ def init_gifti_surfaces_wf(
     return workflow
 
 
+@tag('anat.gifti-morphs')
 def init_gifti_morphometrics_wf(
     *,
     morphometrics: list[str] = ('thickness', 'curv', 'sulc'),
```
effigies, line 943: This one is similarly parameterized.

src/smriprep/workflows/surfaces.py
```diff
@@ -1296,26 +1305,23 @@ def init_anat_ribbon_wf(name='anat_ribbon_wf'):
...
+        (create_pial_distvol, make_ribbon, [('out_file', 'pial_distvols')]),
+        (make_ribbon, outputnode, [('ribbon', 'anat_ribbon')]),
+    ])  # fmt:skip
     return workflow
 
 
+@tag('anat.resample-surfs')
 def init_resample_surfaces_wf(
     surfaces: list[str],
     grayord_density: ty.Literal['91k', '170k'],
```
effigies, line 1327: Again.

effigies (commented):
I guess the idea is to tag anything that might ever be spliced, not things we know need to be?

I'm foreseeing some problems with parameterizable workflows, but those seem unlikely to be spliced. I probably just would not tag them, so that we can worry about how to do it correctly when we need to. Doing it now, we could forget that it needs consideration.

mgxd:
The idea was that the tag could always be overwritten, but I agree it's safer to not have any - a few things to add/consider:
- The splicer function should handle (or complain) if a collision is found
- These multi-use workflows should have some naming convention to make them easy to differentiate

mgxd:
I was thinking something along the lines of `init...pwf` - would be easy to discern and only add a character.

effigies (approved):

PR #477, opened 2025-05-02
REF: generalised anat references 2.0

eilidhmacnicol:
After pushing #475 , @mgxd suggested to move the logic to its own workflow. 

As a result, we have a new function, `collect_anat` which will assign a modality as either 'reference' or 'aux'. At the moment, 'reference' is hard-coded to be T1w but this provides an opportunity for command line overrides or alternative pipelines for nirodents/nibabies

### DIFF
--- a/src/smriprep/utils/misc.py
+++ b/src/smriprep/utils/misc.py
@@ -21,6 +21,7 @@
 #     https://www.nipreps.org/community/licensing/
 #
 """Self-contained utilities to be used within Function nodes."""
+import typing as ty
 
 
 def apply_lut(in_dseg, lut, newpath=None):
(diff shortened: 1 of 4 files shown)


src/smriprep/utils/misc.py
```diff
@@ -95,3 +97,48 @@ def fs_isRunning(subjects_dir, subject_id, mtime_tol=86400, logger=None):
...
+        }
+        for modality in ['t1w', 't2w', 'flair']
+        if modality in subject_data.keys()
+    }
+    anat_inputs[reference_anat.lower()].update(
+        {
+            f'have_{preproc}': f'{reference_anat.lower()}_{preproc}' in precomputed
+            for preproc in ['mask', 'tpms', 'dseg']
+        }
+    )
```
mgxd, line 142: looking at this, i wonder if this precomputed parsing would be better suited for a separate function

mgxd (commented):
had a chance to jot down some thoughts - i like the overall direction, but we can take this moment to clean up the logic (which admittedly is a bit of a mess if derived from nibabies 😅 )

PR #478, opened 2025-05-06
fix: Respect label order in precomputed derivatives

### LINKED ISSUE #474: Incorrect normalized tissue probseg filenames when using precomputed native-space tissue probseg from CAT12
### What happened?

fMRIPrep development team,

I am using precomputed tissue probability maps derived from CAT12 as input to fMRIPrep, to address mislabeling of white matter lesions by FSL FAST as described here (https://neurostars.org/t/handling-of-white-matter-lesions-in-fmriprep/31776).  

However, the normalized probseg files (e.g. sub-X_space-MNI152NLin6Asym_res-2_label-GM_probseg.nii.gz) are mislabeled, such that "label-WM_probseg.nii.gz" contains the grey matter probability map, label-GM contains the CSF map, and label-CSF contains the WM map. I have checked, double-checked, and triple-checked to make sure that the input files in the precomputed derivatives directory are named correctly.  This error seems to be specific to when precomputed derivatives are provided -- when run without them, the files are correctly named.

### What command did you use?

```shell
#!/bin/bash

ml singularity

subj="$1"
singularity run -e --bind /filepath/rawdata:/data \
--bind /filepath/derivatives/functional/fmriprep/fmriprep-25.0.0:/out \
--bind /filepath/pruittpj:/license \
--bind /filepath/derivatives/anatomical/freesurfer:/fsdir \
--bind /filepath/derivatives/functional/fmriprep/sourcedata:/sourcedata \
/filepath/tools/singularity/fmriprep-25.0.0.sif \
/data /out participant \
--participant_label ${subj} --mem 30000 --nprocs 16 --omp-nthreads 8 --dummy-scans 0 \
--output-spaces fsnative fsaverage anat:res-native MNI152NLin6Asym:res-2 --fd-spike-threshold 0.5 --dvars-spike-threshold 3 \
--fs-subjects-dir /fsdir \
--derivatives /sourcedata \
--fs-license-file /license/license.txt \ 
-w /scratch/filepath/fmriprep/
```

### What version of fMRIPrep are you running?

25.0.0

### How are you running fMRIPrep?

Singularity

### Is your data BIDS valid?

Yes

### Are you reusing any previously computed results?

FreeSurfer

### Please copy and paste any relevant log output.

```shell

```

### Additional information / screenshots

When it asks above about whether I was using previously computed results, I'm not sure how to answer -- I am using previously computed anatomical derivatives but not those computed by fMRIPrep

effigies:
`BIDSLayout` does a `natural_sort` on all `get()` query results, which breaks ordering encoded in the query itself. This PR looks for the label entity in the query and re-sorts by the label list to ensure that values are returned as expected.

A more robust refactor might either carry around a list of tissue labels or re-order probsegs, but doing that right would be a nontrivial effort, and introduce opportunities for new bugs. This narrowly targets a known bug.

This is a bug-fix candidate PR.

Fixes #474.

### DIFF
--- a/src/smriprep/utils/bids.py
+++ b/src/smriprep/utils/bids.py
@@ -47,11 +47,17 @@ def collect_derivatives(derivatives_dir, subject_id, std_spaces, spec=None, patt
     derivs_cache = {}
     for key, qry in spec['baseline'].items():
         qry['subject'] = subject_id
-        item = layout.get(return_type='filename', **qry)
+        item = layout.get(**qry)
         if not item:
             continue
 
-        derivs_cache[f't1w_{key}'] = item[0] if len(item) == 1 else item
+        # Respect label order in queries
+        if 'label' in qry:
+            item = sorted(item, key=lambda x: qry['label'].index(x.entities['label']))
+
+        paths = [item.path for item in item]
+
+        derivs_cache[f't1w_{key}'] = paths[0] if len(paths) == 1 else paths
 
     transforms = derivs_cache.setdefault('transforms', {})
     for _space in std_spaces:


effigies:
I've confirmed that the issue is resolved in a local test.

mgxd (approved):
I'm fine with this since it's needed, though this feels like something that should be ultimately addressed in pybids.

effigies:
I don't think pybids should make any particular commitment to an ordering. I'm okay removing the natural sort, or allowing for a sorting parameter, but I think it's best to code defensively on this end...

Issue #429, opened 2023-11-15
Enable longitudinal structural analysis


ChristianNSchmitz:
### What would you like to see added in fMRIPrep?

Dear fMRIPrep-Team,

we are currently working with longitudinal data and would like to use the strucutral freesurfer-output for further analysis. Unfortunately, fMRIPrep creates an average image of all T1 weighted images and only the tweak with the flag --bids-filter-file enables the separate analysis of different sessions. Thus, we would like to ask you if a separate structural analysis of different sessions would be possible. 

Best,
Christian
 

### Do you have any interest in helping implement the feature?

Yes

bpinsard:
@Beatrizfg20 yes, I recently added this feature here https://github.com/nipreps/fmriprep/pull/3142 / https://github.com/nipreps/smriprep/pull/393 to just do that.
I applied this on our dataset, though I used smriprep first steps on the raw T1w/T2w images before feeding these to freesurfer cross-sectional (as it was more robust for skull-stripping), then ran the base and longitudinal templates and provided these to smriprep anat-only to then use it with fmriprep.

tsalo:
@mattcieslak and I are interested in implementing the same option in QSIPrep and we want to ensure that QSIPrep's approach matches sMRIPrep's. Does anyone have any thoughts on what the parameters to enable this would look like?

We had a few ideas:

1. Accept a value for the `--longitudinal` parameter.
    - `--longitudinal register-to-first`: The current default behavior, where everything's registered to the first anatomical scan.
    - `--longitudinal unbiased-template` or `--longitudinal`: How `--longitudinal` currently works, where an unbiased template is generated.
    - `--longitudinal session-wise`: The new proposed behavior, where each session is analyzed separately.
2. A separate parameter that's mutually exclusive with `--longitudinal`, like `--session-wise-anatomical`.
3. Change the behavior of `--longitudinal` to match the new behavior.
4. Enable session-wise anatomical processing by default, like how (I think) Nibabies works.

tsalo:
From today's NiPreps TechMon meeting, the current plan is to (1) deprecate the `--longitudinal` flag in both sMRIPrep and fMRIPrep in the current series and (2) make `--longitudinal` implement session-wise anatomical processing in the next major release. This will be a breaking change.

This plan is predicated on two assumptions AFAICT:

1. Users are not using the `--longitudinal` flag as it is currently implemented.
[...]

PR #479, opened 2025-05-06
Support session-wise anatomical processing

tsalo:
Closes none, but relates to #429.

This is just a draft for now. I think I still need to update the CLI and parser.


mgxd:
Closing in favor of #480

PR #480, opened 2025-05-23
ENH: Integrate per-session processing

mgxd:
Continuation of #479

This breaks the current API so will need to ensure dependencies will be compatible. The plan is to work on this in the fmriprep side once tests are passing.

Remaining TODOs:
- [x] add session level support to `generate_report`


src/smriprep/cli/run.py
```diff
@@ -115,6 +125,14 @@ def get_parser():
         '{<suffix>:{<entity>:<filter>,...},...} '
         '(https://github.com/bids-standard/pybids/blob/master/bids/layout/config/bids.json)',
     )
+    g_bids.add_argument(
+        '--anat-reference-method',
+        choices=['robust-template', 'session'],
```
tsalo, line 130: What about the `first` option? That's the current default.

src/smriprep/cli/run.py
```diff
@@ -127,11 +136,14 @@ def _drop_ses(value):
...
-        '"robust-template" will data across sessions.'
-        '"session" will process each session individually.',
+        choices=['first', 'robust-template', 'session'],
+        default='first',
+        help='Method to produce the reference anatomical space:'
+        '\t"first" will use the first alphabetically sorted image'
+        '\t"robust-template" will construct an unbiased template from all images '
+        '(previously "--longitudinal")'
+        '\t"session" will independently process each session. If multiple sessions are '
+        'found, the behavior will be similar to "first"',
```
tsalo, line 146: How would you feel about `--subject-anatomical-reference {first-alphabetically,unbiased,sessionwise}` instead? That would match QSIPrep's terminology and we wouldn't have to update QSIPrep to match sMRIPrep.

mgxd: I like `unbiased` and `sessionwise`, but `first-alphabetically` feels a bit verbose...if `first` is too vague and you are still open to change, how about `first-abc` or `first-alpha`?

tsalo: `first-alpha` sounds good to me. @mattcieslak WDYT?

src/smriprep/cli/run.py
```diff
@@ -115,6 +134,17 @@ def get_parser():
...
+    g_bids.add_argument(
+        '--subject-anatomical-reference',
+        choices=['first-alpha', 'unbiased', 'sessionwise'],
+        default='first',
+        help='Method to produce the reference anatomical space:'
+        '\t"first-alpha" will use the first alphabetically sorted image'
+        '\t"unbiased" will construct an unbiased template from all images '
+        '(previously "--longitudinal")'
+        '\t"sessionwise" will independently process each session. If multiple sessions are '
+        'found, the behavior will be similar to "first"',
```
tsalo, line 146:
```suggestion
        '\t"sessionwise" will independently process each session. If multiple sessions are '
        'found, the behavior will be similar to "first-alpha"',
```

Wouldn't the behavior match `first-alpha` if _one_ session is found, rather than multiple sessions?

src/smriprep/workflows/base.py
```diff
@@ -174,7 +175,18 @@ def init_smriprep_wf(
...
-    for subject_id in subject_list:
+    for subject_id, session_ids in subject_session_list:
+        # ('01', None) -> sub-01_wf
+        # ('01', 'pre') -> sub-01_ses-pre_wf
+        # ('01', ['pre', 'post']) -> sub-01_ses-pre+post_wf
+
+        name = f'sub-{subject_id}_wf'
+        if session_ids:
+            if isinstance(session_ids, str):
+                ses_str = session_ids
```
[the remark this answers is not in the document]
mgxd: i went back and forth on this - ultimately i decided this could be excessive for studies with many sessions.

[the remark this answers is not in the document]
effigies: IDK, you do still need a way to distinguish if you call with `--session-id A B` in one instance and `--session-id C D` in another with a shared working directory.

```suggestion
            if isinstance(session_ids, str):
                ses_str = session_ids
            else:
                from hashlib import sha1
                ses_str = sha1('+'.join(session_ids).encode()).hexdigest()[:7]
```

src/smriprep/cli/run.py
```diff
@@ -97,6 +109,13 @@ def get_parser():
         help='a space delimited list of participant identifiers or a single '
         'identifier (the sub- prefix can be removed)',
     )
+    g_bids.add_argument(
+        '--session-id',
```
effigies, line 113: Seems worth being consistent with `--participant-label`.

```suggestion
        '--session-label',
```

src/smriprep/cli/run.py
```diff
@@ -53,6 +62,9 @@ def get_parser():
 
     import smriprep
 
+    def _drop_ses(value):
+        return value[4:] if value.startswith('ses-') else value
```
effigies, line 66:
```suggestion
        return value.removeprefix('ses-')
```

pyproject.toml
```diff
@@ -27,6 +27,7 @@ dependencies = [
     "matplotlib >= 3.5",
     "nibabel >= 4.0.1",
     "nipype >= 1.8.5",
+    "nireports @ git+https://github.com/nipreps/nireports.git@main",
```
oesteban, line 30: what's new in nireports that's necessary?

src/smriprep/cli/run.py
```diff
@@ -97,6 +109,13 @@ def get_parser():
         help='a space delimited list of participant identifiers or a single '
         'identifier (the sub- prefix can be removed)',
     )
+    g_bids.add_argument(
+        '--session-label',
+        nargs='+',
+        type=_drop_ses,
+        help='A space delimited list of session identifiers or a single '
+        'identifier (the ses- prefix can be removed)',
+    )
```
oesteban, line 118: How does this interact with bids filters?

src/smriprep/cli/run.py
```diff
@@ -115,6 +134,17 @@ def get_parser():
...
+        '--subject-anatomical-reference',
+        choices=['first-alpha', 'unbiased', 'sessionwise'],
+        default='first',
+        help='Method to produce the reference anatomical space:'
+        '\t"first-alpha" will use the first alphabetically sorted image'
+        '\t"unbiased" will construct an unbiased template from all images '
+        '(previously "--longitudinal")'
+        '\t"sessionwise" will independently process each session. If multiple runs are '
+        'found, the behavior will be similar to "first-alpha"',
+    )
```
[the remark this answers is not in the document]
oesteban: Maybe I'm not understanding "first-alpha" and Chris' suggestion.

I think what's necessary here is a way of designating one particular session as reference right? What would make sense to me is allowing a pattern `time[selector]`:

- `time[0]` selects the session with the T1w image that was collected first as per scans.txt (or acquisition date metadata).
- `time[-1]` selects the last sessioin
- `time[middle]` selects an intermediate point

It also seems `sessionwise` has a consideration for run. unbiased doesn't need it, but how does it work for first-alpha, is the reference then the first run on the first alphabetical session?

src/smriprep/cli/run.py
```diff
@@ -495,6 +528,32 @@ def build_workflow(opts, retval):
...
+            suffix=['T1w', 'T2w'],  # TODO: Track supported modalities globally
+        )
+        if not sessions:
+            if opts.subject_anatomical_reference == 'sessionwise':
+                logger.warning(
+                    '--subject-anatomical-reference "sessionwise" was requested, but no sessions '
+                    f'were found for subject {subject}.'
+                )
+            subject_session_list.append((subject, None))
+            continue
```
oesteban, line 548: I'd rather raise if sessionwise doesn't apply. Also, it should also apply is there's only one session, right?
```suggestion
        if not sessions or len(sessions) == 1 and opts.subject_anatomical_reference == 'sessionwise':
            raise RuntimeError(
                '--subject-anatomical-reference "sessionwise" was requested, but no sessions '
                f'were found for subject {subject}.'
            )
```

src/smriprep/workflows/base.py
```diff
@@ -397,10 +416,15 @@ def init_single_subject_wf(
         run_without_submitting=True,
     )
 
+    if session_id is not None:
+        dismiss_entities = None
+    else:
+        dismiss_entities = ('session',)
```
oesteban, line 422:
```suggestion
    dismiss_entities = None if session_id is not None else ('session', )
```

src/smriprep/workflows/base.py
```diff
@@ -460,13 +483,18 @@ def init_single_subject_wf(
...
+def _prefix(subject_id, session_id=None):
+    """Create FreeSurfer subject ID."""
+    if not subject_id.startswith('sub-'):
+        subject_id = f'sub-{subject_id}'
+    if session_id and not isinstance(session_id, list):
+        # For now, drop ses label if multiple sessions
+        if not session_id.startswith('ses-'):
+            session_id = f'ses-{session_id}'
+        subject_id = f'{subject_id}_{session_id}'
+    return subject_id
```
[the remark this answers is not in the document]
oesteban: perhaps this is a good use-case for the config object? Have it create plain ids or hashes and then just get them from there as a dictionary?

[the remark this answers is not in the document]
mgxd: I'm not concerned about the implementation, but rather the outputs we would produce - imagine a case where `--session-label A B` and `--session-label B C` are run; we can either:
1. not differentiate and leave subject id collisions up to the users
1. hash the sessions and append to subject id

2. sounds better, if we're ok with having outputs like: `sub-X_d9sdf12j`

src/smriprep/cli/run.py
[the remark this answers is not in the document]
effigies: If we're going to have `first-alpha`, I would prefer to change it to `first-lex`, as I think "lexicographic" ordering is both more precise and is readily contrasted with numerical for people who need to look it up.

src/smriprep/workflows/base.py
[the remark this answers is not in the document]
effigies: The odds of collision are low, so we could use fewer characters. What about `ses-multi{hash[:4].upper()}`, so `ses-multi7BE3` for sessions `['01', ..., '20']`?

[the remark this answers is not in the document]
mgxd: Addressed in [c3ee587](https://github.com/nipreps/smriprep/pull/480/commits/c3ee587cae5c03d754a2cf3f3787b415fd9e4dcd) - LMK

src/smriprep/cli/run.py
oesteban: I think we want to fail in this case to be explicit, the behavior in case of a single-session is clear, and this opens it up to having users who ignore the warning and even do not figure out that their dataset is not as they are assuming it is (e.g., because they are just copying the fmriprep command line from a previous project).

IMHO we should keep our stance of not being lenient in these situations. If you want to add a explicit bypass to this exception, I'd be fine (like the bids validation at the beginning -- we fail, but user can deliberately bypass it).

src/smriprep/cli/run.py
```diff
@@ -495,6 +528,32 @@ def build_workflow(opts, retval):
...
+
+        if opts.subject_anatomical_reference == 'sessionwise':
+            for session in sessions:
+                subject_session_list.append((subject, session))
+        else:
+            # This will use all sessions either found by layout or passed in via --session-id
+            # MG: Should session names be concatenated into a label to preserve provenance?
+            subject_session_list.append((subject, sessions))
 
     bids_filters = json.loads(opts.bids_filter_file.read_text()) if opts.bids_filter_file else None
```
oesteban, line 558: I'm concerned this may further filter sessions and change the assumptions done thus far.

src/smriprep/utils/misc.py
```diff
@@ -95,3 +95,37 @@ def fs_isRunning(subjects_dir, subject_id, mtime_tol=86400, logger=None):
     if logger:
         logger.warn(f'Removed "IsRunning*" files found under {subj_dir}')
     return subjects_dir
+
+
+def stringify_sessions(lst: list[str], max_length: int = 10, digest_size: int = 2) -> str:
+    """
+    Convert a list of session into a string identifier.
```
oesteban, line 102:
```suggestion
    Convert a list of session into a string identifier.

```

src/smriprep/utils/misc.py
```diff
@@ -95,3 +95,37 @@ def fs_isRunning(subjects_dir, subject_id, mtime_tol=86400, logger=None):
...
+    'a'
+    >>> stringify_sessions(['a', 'b', 'c'])
+    'a+b+c'
+    >>> stringify_sessions(['a', 'b', 'toolong'])
+    'multi+7c22'
+    >>> stringify_sessions(['a', 'b', 'toolong'], max_length=12)
+    'a+b+toolong'
+    >>> stringify_sessions(['a', 'b', 'toolong'], digest_size=4)
+    'multi+dd8bb349'
+    >>>
```
oesteban, line 120:
```suggestion
```

src/smriprep/cli/run.py
oesteban: Also, my suggestion simplifies the logic, removing one nesting level.

mgxd: i'm fine with starting with an error, and then re-evaluating if it becomes burdensome

src/smriprep/cli/run.py
mgxd: I agree there could be unseen interactions when collecting data, but not sure the best way to account for this without greatly increasing code complexity - we still want to preserve using `--bids-filter-file` in tandem with this session pairing. Open to suggestions

oesteban:
I think it should go the other way around: we should include at least one
basic test with an uncomplicated filter to merge this with the confidence
that at least in a very basic situation both options won't clash. Also, it
seems applying filters should be logically executed before these new
decisions are made in the flow.

oesteban (approved):

PR #482, opened 2025-07-31
ENH: Generate cortex mask in anat_fit_wf

tsalo:
Based on @effigies' idea in https://github.com/nipreps/fmriprep/pull/3491. This new output should make it possible to project data to the surface just using computed derivatives.

Changes proposed:

- Move roi output out of `init_hcp_morphometrics_wf`.
- Calculate that output in a new workflow, `init_cortex_mask_wf`.
- Add step to `init_anat_fit_wf` that calls `init_cortex_mask_wf`.

### DIFF
--- a/src/smriprep/workflows/anatomical.py
+++ b/src/smriprep/workflows/anatomical.py
@@ -1249,12 +1249,12 @@ A {t2w_or_flair} image was used to improve pial surface refinement.
                 ('outputnode.subjects_dir', 'inputnode.subjects_dir'),
                 ('outputnode.fsnative2t1w_xfm', 'inputnode.fsnative2anat_xfm'),
             ]),
-            (gifti_surfaces_wf, surfaces_buffer, [
-                (f'outputnode.{surf}', surf) for surf in surfs
-            ]),
             (sourcefile_buffer, ds_surfaces_wf, [('source_files', 'inputnode.source_files')]),
             (gifti_surfaces_wf, ds_surfaces_wf, [
                 (f'outputnode.{surf}', f'inputnode.{surf}') for surf in surfs
+            ]),
+            (ds_surfaces_wf, surfaces_buffer, [
+                (f'outputnode.{surf}', surf) for surf in surfs
             ]),
         ])
         # fmt:on
@@ -1295,12 +1295,12 @@ A {t2w_or_flair} image was used to improve pial surface refinement.
                 ('outputnode.subject_id', 'inputnode.subject_id'),
                 ('outputnode.subjects_dir', 'inputnode.subjects_dir'),
             ]),
-            (gifti_morph_wf, surfaces_buffer, [
-                (f'outputnode.{metric}', metric) for metric in metrics
-            ]),
             (sourcefile_buffer, ds_morph_wf, [('source_files', 'inputnode.source_files')]),
             (gifti_morph_wf, ds_morph_wf, [
                 (f'outputnode.{metric}', f'inputnode.{metric}') for metric in metrics
+            ]),
+            (ds_morph_wf, surfaces_buffer, [
+                (f'outputnode.{metric}', metric) for metric in metrics
             ]),
         ])
         # fmt:on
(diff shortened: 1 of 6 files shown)


effigies:
e.g.,

```diff
-            (gifti_surfaces_wf, surfaces_buffer, [
-                (f'outputnode.{surf}', surf) for surf in surfs
-            ]),
             (sourcefile_buffer, ds_surfaces_wf, [('source_files', 'inputnode.source_files')]),
             (gifti_surfaces_wf, ds_surfaces_wf, [
                 (f'outputnode.{surf}', f'inputnode.{surf}') for surf in surfs
             ]),
+            (ds_surfaces_wf, surfaces_buffer, [
+                (f'outputnode.{surf}', surf) for surf in surfs
+            ]),
         ])
         # fmt:on
```

src/smriprep/workflows/anatomical.py
```diff
@@ -1344,6 +1346,65 @@ def init_anat_fit_wf(
...
+                extra_entities={'extension': '.label.gii', 'hemi': hemi},
+            )
+
+            workflow.connect([
+                (cortex_mask_wf, ds_cortex_mask_wf, [('outputnode.roi', 'inputnode.mask_file')]),
+                (combine_inputs, ds_cortex_mask_wf, [('out', 'inputnode.source_files')]),
+                (ds_cortex_mask_wf, merge_cortex_masks, [
+                    ('outputnode.mask_file', f'in{i_hemi + 1}'),
+                ]),
+            ])  # fmt:skip
```
mgxd, line 1403: minor gripe: we should aim to keep the high-level workflow as simple as possible - all this wrangling should be contained within the workflow, like `init_hcp_morphometrics`.

mgxd (commented):
personally i prefer `desc-cortex` over `roi`, since the sidecar metadata type is set to roi

src/smriprep/workflows/anatomical.py
effigies: I think the wrangling here is because the stuff that `init_ds_surfaces_wf` does for pairs of surfaces is missing from `init_ds_mask_wf`. I would suggest seeing if we can update `init_ds_surfaces_wf` to handle whatever it lacked from masks. Or just create a hybrid `init_ds_surface_masks_wf`, since it was a special enough case to separate out from volumes.

tsalo: I went with `init_ds_surface_masks_wf`. I think the main workflow should be sufficiently clean now as well.

mgxd (approved):
LGTM

tsalo:
@effigies I think this is ready for one last review. Also, is there anyone else I should ping for a review?

effigies:
@tsalo I wanted to suggest going back to explicit loops, but the changes would be too extensive to make a suggestion without testing. Please see the latest three commits and let me know if you're good with these changes.

src/smriprep/workflows/outputs.py
```diff
@@ -1275,57 +1275,49 @@ def init_ds_surface_masks_wf(
...
+    workflow.connect([
+        (inputnode, select_files, [
+            ('mask_files', 'mask_file'),
+            ('source_files', 'sources'),
+        ]),
+        (ds_itersource, select_files, [('hemi', 'key')]),
+        (ds_itersource, ds_surf_mask, [('hemi', 'hemi')]),
+        (select_files, ds_surf_mask, [
+            ('mask_file', 'in_file'),
+            ('sources', 'Sources'),
```
tsalo, line 1316: This won't be a BIDS relative path anymore though, right?

effigies: It already wasn't:

https://output.circle-artifacts.com/output/job/ad802373-bcd6-4afc-b242-4a701053a959/artifacts/0/fasttrack/smriprep/sub-01/anat/sub-01_hemi-L_desc-cortex_mask.json

```json
{
  "Sources": [
    "/tmp/ds005/derivatives/smriprep/sub-01/anat/sub-01_hemi-L_midthickness.surf.gii",
    "/tmp/ds005/derivatives/smriprep/sub-01/anat/sub-01_hemi-L_thickness.shape.gii"
  ],
  "Type": "ROI"
}
```

PR #485, opened 2025-08-06
chore(ci): Run pytest with -n auto on CircleCI

effigies:
Hopefully this will make the pytest job go faster.


mgxd (commented):
not sure i'm reading this right... somehow this made the tests take longer?

effigies:
I guess we can revert the ds0* ones. I don't know why ds005 slowed down. But at least pytest sped up.

PR #497, opened 2025-09-24
fix: Use buffer node to hold fsnative2anat transform

effigies:
When a precomputed fsnative2anat transform was found, we were not recomputing it, but it was not injected into the workflow. The bug could be considered to be part of failing to set the `outputnode` of the `surface_recon_wf`, but it feels easier to me to keep track of in the top-level fit workflow.

@tsalo Can you patch this into your container?


effigies:
@mgxd Tests are passing (ignore pre-commit). Do you agree that this should be a bug-fix? A new buffernode changes the in-memory workflow, but shouldn't add any new execution nodes, so this should not invalidate a scratch directory.

PR #526, opened 2026-01-07
Rename space output to template and add new space output to template_iterator_wf

tsalo:
Related to https://github.com/nipreps/nibabies/issues/501 and https://github.com/nipreps/fmriprep/issues/3365.

Changes proposed:
- Modify `space` output in `init_template_iterator_wf` to be the space entity formatted as `<space>+<cohort>`.
- Add new `template` output to `init_template_iterator_wf` to track the space-only version of space (which is what the template entity should have).
- Remove `cohort` input to `init_ds_anat_volumes_wf`. This will change anatomical output files from `<source_entities>_space-<space>_cohort-<cohort>_...` to `<source_entities>_space-<space>+<cohort>_...`.


mgxd (approved):
LGTM as is, but small nit is since we now have very similar nodes in `gen_tplid` and `gen_space_entity` it might make sense to combine them into a dedicated interface.

mgxd (approved):

PR #580, opened 2026-08-21
feat: Validate subject directory exists if --fs-no-resume

effigies:
This addresses nipreps/fmriprep#3669. The root cause of that issue was that a sessionwise FreeSurfer subject ID was being generated and searched for in an existing FreeSurfer subjects directory that used only the participant label. When combined with `--fs-no-resume`, this led to searching for files in the existing directory, finding none, and then passing empty lists to nodes that expected existing files.

This replaces the resulting inscrutable errors with a `RuntimeError` that clearly indicates the subject ID that was expected. Unfortunately, this happens at workflow runtime, but moving to workflow construction time would be too much of a modification for a patch release.

I believe this is worth bending the patch-release rules, which generally require no changes to workflows. For successful workflows, rerunning should only rerun validation, as it simply passes the values through. The gain in an interpretable error message in an LTS series is worth it, IMO.

@mgxd Do you agree?

### DIFF
--- a/src/smriprep/workflows/surfaces.py
+++ b/src/smriprep/workflows/surfaces.py
@@ -58,7 +58,7 @@ import smriprep
 from smriprep.interfaces.surf import MakeRibbon
 from smriprep.interfaces.workbench import SurfaceResample
 
-from ..interfaces.freesurfer import MakeMidthickness, ReconAll
+from ..interfaces.freesurfer import MakeMidthickness, MRIsConvertData, ReconAll, ValidateSubjectDir
 from ..interfaces.gifti import MetricMath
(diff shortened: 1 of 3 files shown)


mgxd (approved):
LGTM, I agree this is a case where I'm okay with a imperfect backport

src/smriprep/workflows/surfaces.py
```diff
@@ -286,11 +286,16 @@ def init_surface_recon_wf(
             ]),
         ])  # fmt:skip
     else:
+        validate_subject_dir = ValidateSubjectDir()
```
effigies:
```suggestion
        validate_subject_dir = pe.Node(ValidateSubjectDir(), name='validate_subject_dir')
```

PR #585, opened 2026-09-08
rf: Transition to nipost.collect_derivatives

effigies:
Pins nipreps/nipost#1, and adjusts data retrieval code.


src/smriprep/workflows/anatomical.py
```diff
@@ -1352,12 +1354,12 @@ def init_anat_fit_wf(
...
     elif msm_sulc:
         LOGGER.info('ANAT Stage 10: Found pre-computed MSM-Sulc registration sphere')
-        msm_buffer.inputs.sphere_reg_msm = sorted(precomputed['sphere_reg_msm'])
+        msm_buffer.inputs.sphere_reg_msm = sphere_reg_msm
     else:
         LOGGER.info('ANAT Stage 10: MSM-Sulc disabled')
 
     # Stage 11: Cortical surface mask
-    if len(precomputed.get('cortex_mask', [])) < 2:
+    if not (cortex_mask := surfaces.get('cortex_mask')):
```
mgxd: i believe this should be
```suggestion
    if not (cortex_mask := surfaces.get('cortex')):
```

come to think of it, since we're reworking this it would be nice to have a stricter catch.

effigies: What is "stricter catch"?

mgxd: an error - perhaps `surfaces` should be pre-populated, and indexed via brackets

effigies: Sounds good. Let me transition the nipost collector to always return populated fields with `None`s for failing matches.

